PR #24, opened 2020-12-15
Add FFTW, Bison, M4, Flex, readline, ncurses, Abseil, Google test, NGSPICE, ABC, Netgen, Magic, Icarus Verilog, Icestorm

per-gron:
They are dependencies of NGSPICE which I intend to add in a follow up commit.


WORKSPACE
```diff
@@ -12,6 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+workspace(name = "com_google_bazel_rules_hdl")
```
mithro, line 15: Should this be something like `com_github_hdl_bazel_rules_hdl`?

dependency_support/com_github_westes_flex/bundled.BUILD.bazel
```diff
@@ -0,0 +1,156 @@
+# Copyright 2020 Google LLC
```
mithro, line 1: It's annoying that GitHub doesn't seem to syntax highlight this file :-(

mithro (commented):
Two random comments.

dependency_support/com_github_westes_flex/bundled.BUILD.bazel
umarcor: This is a very blind guess, but maybe using `.gitattributes` as in https://github.com/im-tomu/fomu-workshop/pull/390 does the trick?

per-gron: I tried adding `.gitattributes` but it doesn't seem to have an effect. Maybe it needs to be merged first?

WORKSPACE
per-gron: As per private chat, I'm changing it to `rules_hdl`

dependency_support/com_github_westes_flex/bundled.BUILD.bazel
umarcor: Effective or not, it is harmless and it's always good to be explicit about these "metadetails". Hence, let's leave it.

Anyway, linguist is meant for the repo language usage statistics. The effect on syntax hightlighting was just something I hoped that might work.

QuantamHD (approved):

per-gron:
@QuantamHD I rebased the PR and fixed the conflicts. PTAL.

per-gron:
I added a bunch of more things. Sorry about the huge PR... But in this case I think splitting it up doesn't really make it more reviewable since it's all quite separate.

per-gron:
Once this is reviewed I can continue with Tcl, Yosys, Project Trellis, Nextpnr. That's most of what I have I think.

dependency_support/org_gnu_m4/bundled.BUILD.bazel
```diff
@@ -0,0 +1,128 @@
...
+        "lib/config.h",
+        "lib/configmake.h",
+        "build-aux/snippet/unused-parameter.h",
+    ] + HEADERS_WITH_C_EXTENSION,
+    includes = ["lib", "build-aux/snippet"],
+    visibility = ["//visibility:private"],
+)
+
+cc_binary(
+    name = "m4",
```
hzeller, line 54: Question: would this manual m4 building work with clang/clang++ as well ?

In [verible](https://github.com/google/verible), we build m4 mentioned in our [WORKSPACE file](https://github.com/google/verible/blob/master/WORKSPACE#L99-L107), using plain [configure_make()](https://github.com/google/verible/blob/master/bazel/BUILD#L20-L29).

However, when building with clang:
```
git clone https://github.com/google/verible.git && cd verible
CC=clang CXX=clang++ bazel build //bazel:m4
```
we run into compile errors apparently having to do with an undefined symbol used by clang's code-generation ?
```
../lib/libm4.a(xmalloc.o):xmalloc.c:function xnmalloc: error: undefined reference to '__muloti4'
../lib/libm4.a(xmalloc.o):xmalloc.c:function xnrealloc: error: undefined reference to '__muloti4'
../lib/libm4.a(xmalloc.o):xmalloc.c:function xcalloc: error: undefined reference to '__muloti4'
```

This means, that verible currently can **not** be built on MacOS (where clang is the default compiler), which limits the audience; So if there is a way to make m4 build reliably on that platform as well, we  (@fangism and me) would be very happy to accept guidance (could be in the shape of a pull request :D) how to make this work in verible.

With regard to this pull request, it probably makes sense to test all of these new build rules also with `CC=clang CXX=clang++` to make sure we're somewhat compiler-independent.

hzeller (approved):
LGTM

dependency_support/org_gnu_m4/bundled.BUILD.bazel
per-gron: The BUILD rules in this PR are all copied from a repository that uses clang as the compiler, so the answer is yes but with this caveat:

It works only if the m4 binary is linked with LLVM's compiler-rt (as opposed to libgcc). See https://github.com/android/ndk/issues/295#issuecomment-277098658 and the linked LLVM issue https://bugs.llvm.org/show_bug.cgi?id=16404. The main linker flag you need is `-rtlib=compiler-rt`. A quick way to get it to work is to build with `CC=clang CXX=clang++ bazel build --linkopt=-rtlib=compiler-rt @org_gnu_m4//:m4`. Another nice and easy way of setting up a clang toolchain is to use [this LLVM Bazel toolchain](https://github.com/grailbio/bazel-toolchain/tree/2838e04e85076ad96edb19c3412d68069d41bccb) which easily gives you a nearly (it doesn't disable global system header include paths) hermetic toolchain.

per-gron:
I don't have merge access to this repo, so if someone can push the merge button that'd be good. Then I'd be able to make a fresh follow-up PR without adding to this insanely large one :smile:

dependency_support/org_gnu_m4/bundled.BUILD.bazel
per-gron: Once I've pulled in the deps that I'm working on now I'd be happy to set up CI that runs Clang and GCC. @mithro could maybe pull some strings to give me access to a GCP project that has Cloud Build enabled? I have a feeling that the CI solutions that can be had for free for FOSS projects are not quite going to be fast enough to be nice to use.

hzeller: Is there a way in a bazel BUILD file to automatically enable `compiler-rt` depending on the toolchain used ?

mithro:
@per-gron - Your access level should now be fixed.

dependency_support/org_gnu_m4/bundled.BUILD.bazel
per-gron: @hzeller Last time I checked Bazel did not offer a way to `select` based on GCC vs Clang. It's really weird to me that they don't offer this, it's really quite common to have to choose copts based on this. I think you could `select` based on the OS, see https://github.com/tensorflow/tensorflow/blob/master/third_party/clog/BUILD.bazel

Since this is arguably a bug in the toolchain the "proper" way to fix it (short of fixing LLVM) is to change the Bazel toolchain configuration, which is what using https://github.com/grailbio/bazel-toolchain does for you, although that also downloads LLVM binaries.

PR #27, opened 2020-12-17
Add Tcl, libffi, Yosys, Project Trellis, Boost, Eigen and Nextpnr

per-gron:
This should be all external dependencies I will add for now.

Project Trellis uses C++17 so I added a `.bazelrc` that configures the whole project to use that.


per-gron:
I found that Project Trellis in this PR doesn't work on my workstation, I'll close the PR and try to fix

PR #29, opened 2020-12-20
Adds Skywater rules and PDK provider data.

QuantamHD:
These rules allow subsequent sythesis rules to consume PDK data through a stadard provider API.


pdk/BUILD
```diff
@@ -0,0 +1 @@
+# Required to be a package
```
per-gron, line 1: I think we should make sure to have the Apache license preamble in every file (this comment applies to other files as well)

pdk/skywater130/build_defs.bzl
```diff
@@ -0,0 +1,103 @@
+"""Skywater 130 PDK support rules.
```
per-gron, line 1: What's the rationale of having this file here?

I think the Skywater PDK related files can be under `pdk/skywater130`, or it could be under `dependency_support/...`. But this structure, where there are dependencies from `dependency_support` into the main `rules_hdl` workspace doesn't make sense to me because then there are dependencies from what's supposed to be the dependencies of this repo to the repo itself. It feels very circular to me even though it's technically not.

per-gron: In general I'm a big fan of keeping related (as in providing a certain feature) files in the same place. One reason I'm asking this is because it's not entirely clear what the boundary is that explains why the Skywater PDK should be in two places.

QuantamHD:
@per-gron Please take a look. It looks like the hdl-bazel-rules-hdl-push (foss-fpga-tools-external) is running into some issues.

per-gron:
@QuantamHD The CI is fixed by #32. But I will also "fix" this by making hdl-bazel-rules-hdl-push only trigger for changes to the main branch, there is no need to run both a check because you created a PR and because you pushed to a branch.

I didn't encounter this problem because my PR was from a branch in my fork.

dependency_support/com_google_skywater_pdk/build_defs.bzl
```diff
@@ -0,0 +1,121 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Skywater 130 PDK support rules.
+
+These rules
```
per-gron, line 17: This sentence is incomplete.

dependency_support/com_google_skywater_pdk/declare_cell_library.bzl
```diff
@@ -42,25 +35,22 @@ def declare_cell_library(library_name):
...
+        skywater_corner(
+            name = "{}".format(corner),
+            visibility = ["//visibility:private"],
+            srcs = native.glob([
+                "cells/**/*.lib.json",
+                "timing/*.lib.json",
+            ]),
+            standard_cell_name = library_name,
+            ccsnoise = ccsnoise,
+            leakage = leakage,
```
per-gron, line 49: Optional/nit: For `ccsnoise` and `leakage` it's not obvious from the names that these are booleans that decide whether they should be generated or not. Maybe `with_ccsnoise` or `generate_ccsnoise` would be clearer names?

pdk/build_defs.bzl
```diff
@@ -0,0 +1,19 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Providers for PDKs to be used by downstream synthesis.
+"""
+
+StandardCellInfo = provider("Contains information about the standard cells used for synthesis", fields = ["corners", "default_corner"])
+CornerInfo = provider("Contains information about the standard cells used for synthesis", fields = ["liberty", "ccsnoise", "leakage", "corner_name"])
```
per-gron, line 19: If these are intended to be reusable I think it could help to document what these fields are, at least what type they have. (Are ccsnoise and leakage booleans here too?)

per-gron (approved):
Minor comments about code clarity, but otherwise LGTM, feel free to merge when you think is appropriate.

PR #32, opened 2021-01-15
Fix hdl-bazel-rules-hdl-push CI configuration

per-gron:
Before this PR hdl-bazel-rules-hdl-pr works, but not hdl-bazel-rules-hdl-push.

### DIFF
--- a/cloudbuild.yaml
+++ b/cloudbuild.yaml
@@ -60,4 +60,4 @@ steps:
     path: /root/.ssh
 
 substitutions:
-  _USERNAME: $(pull_request.pull_request.user.login)
+  _USERNAME: $(commit.author.login)


umarcor:
Maybe both are required, depending on the trigger event type (and associated metadata)?

per-gron:
> Maybe both are required, depending on the trigger event type (and associated metadata)?

@umarcor The username that this commit is about is only used in one way: It's sent to Buildbuddy, which uses this to say "Per-gron's build //... [...]" in the heading of the build log. See https://app.buildbuddy.io/invocation/7c57127a-572e-11eb-a08d-0242c0a80a02 Without this it says "root's build ..." which I thought was ugly/confusing.

umarcor:
@per-gron, it's ok. On GitHub Actions sometimes you need to get these fields differently depending on the event (a push, a PR, scheduled, etc.). I just wanted to check that you were not unwantedly fixing one case but breaking the other (by making you think it quickly a second time :wink:).

PR #34, opened 2021-01-21
Enable Bazel cache backed by Google Cloud Storage

per-gron:
I tried using Buildbuddy for this, but uploads of some large artifacts time out and fail the whole build. Hopefully GCS will perform better here, either by not having the timeout, or by being faster, let's see.

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -17,6 +17,9 @@ build --host_cxxopt "-std=c++17"
 
 build:ciremotebuild --crosstool_top=@llvm_toolchain//:toolchain
 build:ciremotebuild --bes_backend=grpcs://cloud.buildbuddy.io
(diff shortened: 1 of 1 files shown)


per-gron:
Build time with cold cache: 49 minutes

mithro:
@kgugala has been looking at using GitHub Actions self hosted runners on Google Cloud with throw away support (so can be used on public repos).

per-gron:
I didn't set up ACLs correctly "Anonymous caller does not have storage.objects.get access to the Google Cloud Storage object."

per-gron:
Ok I think this works now. With a warm cache the build time is 27 mins, a nice ~1/2 reduction for 3 LOC.

PR #38, opened 2021-06-02
Fix tk_tcl COPTS escape, bump to abseil/abseil-cpp@278e0a0, YosysHQ/yosys@de79978

jesec:


QuantamHD:
Thanks for sending the PR. Will approve if the CI passes

PR #43, opened 2021-06-26
Simplify m4 and bison rules with "org_gnu_gnulib" and fixup "tk_tcl" build

jesec:
Can be used by GNU tools like m4 and bison.

This change consolidates the portability and Bazel works of GNU tools
into a unified place, in effect offloads the HARD parts from them.
Hopefully, this will make it easier for us to maintain rules of those
tools.

Plus, this version of gnulib is fairly recent, which provides much
better compatibility.

Most of the rule comes from https://github.com/jmillikin/rules_m4,
which is licensed under Apache 2.0.


jesec:
```
error running 'git fetch origin refs/heads/*:refs/remotes/origin/* refs/tags/*:refs/tags/*' while working with @org_tuxfamily_eigen:
error: RPC failed; curl 56 GnuTLS recv error (-54): Error in the pull function.
fatal: The remote end hung up unexpectedly
fatal: early EOF
fatal: index-pack failed
```

Well. Not my fault. Is there a way to rerun? 

/gcbrun

?

hzeller:
Did you verify that bison can be built under MacOS ? In verible, we once [switched to](https://github.com/google/verible/pull/728) the jmillikin rules to get a working compilation across platforms (as there were otherwise all kinds of troubles with gnulib and m4); just want to make sure that these jmillikin-derived rules preserve that property.

jesec:
> Did you verify that bison can be built under MacOS ? In verible, we once [switched to](https://github.com/google/verible/pull/728) the jmillikin rules to get a working compilation across platforms (as there were otherwise all kinds of troubles with gnulib and m4); just want to make sure that these jmillikin-derived rules preserve that property.

At the moment, I can't say for sure about macOS support with this PR. 

However, I am fairly sure that the rules (w/o this PR) do not work for macOS right now. This PR brings substantial portability improvement, and we can certainly try to further improve to support macOS in the future.

jesec:
I manage to test on a macOS machine. With this PR, it is indeed possible to compile working `bison` and `m4` executable on macOS. 

image

hzeller (approved):
Nice!
For the question of portability, maybe for later we can consider CIs that run on a mac/windows environment.

jesec:
Bump! 

Can someone 🚢 this?

mithro:
Should this be merged? @QuantamHD

QuantamHD:
Sorry for the late reply @jesec merging now.

PR #44, opened 2021-08-04
com_github_westes_flex, org_gnu_gnulib: portability improvements

jesec:
**com_github_westes_flex: drop unnecessary locale bits**

Dropped `ENABLE_NLS`, `HAVE_GETTEXT` and other locale bits from `defs`. 

```
Users most likely won't benefit from localized messages.

This change removes an unwanted dependency on locale libs of the
system, and makes "flex" compile on platforms that don't have one
and platforms that have different-than-expected one.
```

**org_gnu_gnulib: add header declarations of obstack for macOS**

Fixes `implicit declaration of function` errors on macOS.

### DIFF
--- a/dependency_support/com_github_westes_flex/bundled.BUILD.bazel
+++ b/dependency_support/com_github_westes_flex/bundled.BUILD.bazel
@@ -12,15 +12,18 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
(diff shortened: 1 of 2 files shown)


mithro:
@QuantamHD - PTAL.

QuantamHD:
Thanks so much for the PR's @jesec these are awesome!!!

PR #45, opened 2021-08-14
Synthesis: Added the ability to use yosys to synthesize verilog_library rules

QuantamHD:
This PR introduces two major concepts
	* verilog_library and its associated providers.
	* synthesize_rtl which takes a PDK and generates a synthesized
	netlist.

Both rules generate consumable providers that can be used by downstream
rules.


mithro:
At some point this will probably need tests?

QuantamHD:
The test is supposed to be this file. During CI we build and test all of the targets. If not synthesized verilog is produced that's essentially the test. I'm a little bit worried about asserting anything about the synthesized output. We could potentially run LVS on the two netlists, but I'm worried because that seems more like a test for yosys rather than a test for the rules.

PR #48, opened 2021-09-03
depedency_support: Adding LEMON and GLPK

QuantamHD:
These dependencies solve SAT type constraint problems.
LEMON is a depednecy of OpenROAD.


hzeller (approved):
LGTM

dependency_support/com_github_quantamhd_lemon/lemon.patch
```diff
@@ -0,0 +1,2945 @@
...
++#include "lemon/counter.h"
++#include "lemon/suurballe.h"
++#include "lemon/graph_to_eps.h"
++#include "lemon/lgf_writer.h"
++#include "lemon/arg_parser.h"
++#include "lemon/euler.h"
++#include "lemon/math.h"
++#include "lemon/kruskal.h"
++#include "lemon/time_measure.h"
+
```
per-gron, line 2934: Sorry @QuantamHD about the slow review. This kind of issue I have otherwise resolved by using either the `includes` attribute of `cc_library`, or if that doesn't work `copts`. No need to change that now, but a patch like this is probably going to be a bit of a pain to maintain with version bumps.

QuantamHD: So the issue is that bazel doesn't allow you to `includes` angle includes. The copts solution would work, but it has the disadvantage that all future importers have to also include those copts in their cc_* rules. Also this library hasn't been updated in over 7 years (2014-07-07) so I'm not too concerned about the patch size.

per-gron: I misremembered: It is `include_prefix` and `strip_include_prefix` that you can (hackily) use to work around this. If they are used it's possible to import with `<>`.

per-gron: But it seems fine to leave this as is :smile:

PR #49, opened 2021-09-05
dependency_support: Added swig 4.0.0 to the repo.

QuantamHD:
This is required for OpenROAD to generate its swig bindings.


dependency_support/dependency_support.bzl
```diff
@@ -37,6 +37,7 @@ load("//dependency_support/com_opencircuitdesign_netgen:com_opencircuitdesign_ne
 load("//dependency_support/edu_berkeley_abc:edu_berkeley_abc.bzl", "edu_berkeley_abc")
 load("//dependency_support/net_invisible_island_ncurses:net_invisible_island_ncurses.bzl", "net_invisible_island_ncurses")
 load("//dependency_support/net_sourceforge_ngspice:net_sourceforge_ngspice.bzl", "net_sourceforge_ngspice")
+load("//dependency_support/net_sourceforge_prdownloads_swig:net_sourceforge_prdownloads_swig.bzl", "net_sourceforge_prdownloads_swig")
```
per-gron, line 40: Why `net_sourceforge_prdownloads_swig` and not `org_swig`?

dependency_support/net_sourceforge_prdownloads_swig/bundled.BUILD.bazel
```diff
@@ -0,0 +1,911 @@
...
+    ],
+    features = [
+        "-parse_headers",
+        "-use_header_modules",  # Incompatible with -fexceptions.
+    ],
+    includes = includes,
+    output_licenses = ["unencumbered"],
+    deps = [
+        "@org_pcre_ftp//:pcre",
+    ],
```
per-gron, line 910: Should this have public visibility?

per-gron (approved):

dependency_support/dependency_support.bzl
per-gron: Workspace names like this are a part of the public API of the package: Its name will be mentioned in all of the build files of all of the projects that use swig via rules_hdl. For this reason, changing a workspace name is difficult or impossible to do.

I think `org_swig` is a better choice here, because if SWIG would change where they host their files, it doesn't make sense that we need to break our public API for that reason. What matters to the user is that this package is SWIG, not where it was downloaded from. I have used net_sourceforce_* as a prefix only when there was no dedicate domain name to use.

PR #59, opened 2021-11-02
Update pybind11_bazel and use that to switch to hermetic Python.

nanavati:


QuantamHD:
LGTM!!! Thanks for the PR

PR #60, opened 2021-11-10
Add self-contained tar files for openroad and yosys

nanavati:


WORKSPACE
```diff
@@ -84,6 +84,22 @@ maybe(
...
+        "https://mirror.bazel.build/github.com/bazelbuild/rules_pkg/releases/download/0.5.1/rules_pkg-0.5.1.tar.gz",
+        "https://github.com/bazelbuild/rules_pkg/releases/download/0.5.1/rules_pkg-0.5.1.tar.gz",
+    ],
+    sha256 = "a89e203d3cf264e564fcb96b6e06dd70bc0557356eb48400ce4b5d97c2c3720d",
+)
+
+load("@rules_pkg//:deps.bzl", "rules_pkg_dependencies")
+rules_pkg_dependencies()
+
+load("@rules_pkg//toolchains:rpmbuild.bzl", "rpmbuild_register_toolchains")
```
QuantamHD, line 100: Do we need the rpm package rules?

WORKSPACE
```diff
@@ -84,6 +84,22 @@ maybe(
...
+    name = "rules_pkg",
+    urls = [
+        "https://mirror.bazel.build/github.com/bazelbuild/rules_pkg/releases/download/0.5.1/rules_pkg-0.5.1.tar.gz",
+        "https://github.com/bazelbuild/rules_pkg/releases/download/0.5.1/rules_pkg-0.5.1.tar.gz",
+    ],
+    sha256 = "a89e203d3cf264e564fcb96b6e06dd70bc0557356eb48400ce4b5d97c2c3720d",
+)
+
+load("@rules_pkg//:deps.bzl", "rules_pkg_dependencies")
+rules_pkg_dependencies()
```
QuantamHD, line 98: Seems like this should go in dependency_support

WORKSPACE
nanavati: Not specifically, but `bazel test @rules_pkg://...` blows up if an rpmbuild toolchain isn't registered.

WORKSPACE
nanavati: I tried. The problem I had is that I couldn't get `bazel test @rules_pkg//...` to work (at least locally) without registering an rpmbuild toolchain at the top level (which pushed the `rules_pkg` repo to the top level so you can load `rpmbuild_register_toolchains`. If you can think of a different way to deal with that, I'm all ears.

nanavati:
Unfortunately rules_pkg has a dependency on the JSON API that was added in Bazel 4 that I can't find a way to work around (so it doesn't work with the bazel 3.5 in the CI).

I think this means that we have to use the buggy, deprecated and explicitly unsupported pkg_tar support for now. I'll make a new pull request with that (which appears to work if you ignore a trap where they generate unusable tar files), but this whole experience makes me think that paying off the old-Bazel technical debt we have is more important than I previously realized.

QuantamHD:
Okay let me reach out to my contact in the bazel org.

PR #64, opened 2021-11-15
Change the synthesize_rtl rule to use a tcl script as an argument

nanavati:


synthesis/build_defs.bzl
```diff
@@ -118,5 +103,9 @@ synthesize_rtl = rule(
             executable = True,
             cfg = "exec",
         ),
+        "synth_tcl": attr.label(
```
QuantamHD, line 106: We should add a doc attribute to this parameter explaining the high level API of the environment variables.

QuantamHD:
This is very clean! I like the new approach.

nanavati:
I think this is ready-to-go (unless you think the doc needs to be clearer or have more detail)

PR #66, opened 2021-11-24
Switch to the rules_pkg packaging rules after the Bazel 4 upgrade

nanavati:


QuantamHD:
Can you update the README with the required bazel version?

cloudbuild.yaml
```diff
@@ -33,7 +33,7 @@ steps:
     path: /root/.ssh
 
 # Run the build
-- name: l.gcr.io/google/bazel:3.5.0
+- name: l.gcr.io/google/bazel:4.2.1
```
QuantamHD, line 36: There's no such thing as l.gcr.io/google/bazel:4.2.1

The highest it goes today 3.5. That's why I install it manually below.

nanavati: Yeah, I thought it was just an internal name and discovered I was wrong. Fixed.

PR #71, opened 2021-12-07
Add `skip_detailed_routing` bool attribute to `place_and_route` bzl

grotival:
detailed_routing takes the bulk of the execution time but is only required for a fully manufacturable chip.
When skip_detailed_routing == True, the open_road_provider.routed_def is not present in the output.


QuantamHD:
LGTM! Thanks for the change.

PR #75, opened 2021-12-21
at_clifford_yosys: replace ternary "-D" with patch, ncurses: cp with "-f" to deal with unsandboxed cases 

### LINKED ISSUE #73: Compiler -D argument escape issues with newer Bazel (gcc: error: .: No such file or directory)
Bazel 4.2.2, GCC 9.3.0

at_clifford_yosys:

```
gcc: error: ?: No such file or directory
gcc: error: std::getenv("YOSYS_DATDIR"): No such file or directory
gcc: error: :: No such file or directory
gcc: error: ".": No such file or directory
```

arguments: `'-DYOSYS_SRC="./"' '-DYOSYS_DATDIR=std::getenv("YOSYS_DATDIR") ? std::getenv("YOSYS_DATDIR") : "."'`

source: https://github.com/hdl/bazel_rules_hdl/blob/9192cbcd04fee629492f906fb443c4bc760e1745/dependency_support/at_clifford_yosys/bundled.BUILD.bazel#L129-L130

jesec:
at_clifford_yosys: replace ternary "-D" with patch (fix #73):

```
There seems to be a regression (or behavior change) with Bazel 4.2.2 in string escaping or copts sanitization.
```

ncurses: cp with "-f" to deal with unsandboxed cases:

```
Fixes: "cp: cannot create regular file './make_hash': File exists"

This issue only happens when Bazel runs with unsandboxed strategies
like "local" or "standalone".

On some platforms (e.g. macOS), debug symbols of compiled binaries
are stored in separate file(s). Bazel fails to preserve those files
in sandboxed mode (default). In other cases, we may wish to inspect
the intermediate files generated during an action. As such, it is
often necessary to run Bazel with unsandboxed strategies.
```

### DIFF
--- a/dependency_support/at_clifford_yosys/at_clifford_yosys.bzl
+++ b/dependency_support/at_clifford_yosys/at_clifford_yosys.bzl
@@ -27,4 +27,7 @@ def at_clifford_yosys():
         strip_prefix = "yosys-95c60866813e520da48c628d4f98a2fe2cb4db25",
         sha256 = "bdeaabd03db2c90b040edc6c6616b573f7712224a0ff7722c8f6c9cf2d490431",
         build_file = Label("@rules_hdl//dependency_support:at_clifford_yosys/bundled.BUILD.bazel"),
+        patches = [
+            Label("@rules_hdl//dependency_support:at_clifford_yosys/yosys.patch"),
+        ],
     )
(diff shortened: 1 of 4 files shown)


QuantamHD:
LGTM! Thanks for the change as always @jesec

PR #82, opened 2022-01-04
Eval power & area at various levels:

grotival:
Eval power & area at various levels:
1. on Verilog
2. after general routing


QuantamHD:
LGTM!

PR #84, opened 2022-01-05
Remove redundant clock creation from place and route scripts.

nanavati:


place_and_route/private/clock_tree_synthesis.bzl
```diff
@@ -46,9 +46,6 @@ def clock_tree_synthesis(ctx, open_road_info):
             liberty_file = liberty.path,
         ),
         "source {}".format(rc_script.path) if rc_script else "",
-        "create_clock [get_ports clk] -period {period}".format(
```
QuantamHD, line 49: Is this actually redundant? My understanding was that this information isn't saved across db saves.

nanavati: I'm pretty sure it is (and I think synthesis worked fine without repeating the clock info).

create_clock is effectively a shortcut for a very simple sdc file. The flows I've seen externally seem to expect that constraints from an sdc file are persisted in the db, and create_clock shouldn't be different.

PR #88, opened 2022-02-03
Add a rule for creating single-file Yosys synthesis steps

nanavati:


QuantamHD:
Looks like there's a lint failure/

flows/yosys/synth.tcl
```diff
@@ -0,0 +1,30 @@
...
+
+# generic synthesis
+set top $::env(CONSTANT_TOP)
+yosys synth -top $top
+
+# mapping to liberty
+set liberty $::env(LIBERTY)
+dfflibmap -liberty $liberty
+
+#if { [info exists ::env(CLOCK_PERIOD) ] } {
```
vincent-mirian-google, line 19: Maybe avoid commented code?

flows/flows.bzl
```diff
@@ -215,6 +215,13 @@ def _run_step_with_inputs(ctx, step, inputs_dict, outputs_step_dict):
 
     return dicts.add(inputs_dict, outputs_dict)
 
+def target_to_file(target):
+    all_files = target.files.to_list()
+    if len(all_files) != 1:
+        fail("Input target", target, "should provide exactly one file instead of", all_files)
```
vincent-mirian-google, line 221: Is a space needed between the quoted values and the variables?... "Input target "?
Should you mention the length or did you want to list the files?

flows/yosys/synth.tcl
nanavati: Fixed (at least partially, as noted in the TODO it isn't tested completely)

vincent-mirian-google:
LGTM! Thanks.

flows/yosys/build_defs.bzl
```diff
@@ -0,0 +1,97 @@
...
+
+    liberty = standard_cells.default_corner.liberty
+
+    synth_tcl = ctx.file.synth_tcl
+
+    cell_runfiles = ctx.runfiles(files = [synth_tcl, liberty, yosys_wrapper])
+
+    yosys_args = [
+        # "-q", # quiet mode only errors printed to stderr
+        # "-q", # second q don't print warnings
```
QuantamHD, line 36: Do you really want all this info to be printed in shell? I don't feel like that should be the default. @cdleary had previously filed a bug about that.

flows/yosys/synth.tcl
```diff
@@ -0,0 +1,30 @@
+# Default Yosys synthesis tcl script for the yosys_synth_file_step rule.
+# It can be replaced by a user-defined script by overriding the synth_tcl
+# argument of that rule.
+
+yosys -import
+
+# read design
+set input_rtl $::env(INPUT_RTL)
+yosys read_verilog -sv -defer $input_rtl
```
QuantamHD, line 9: Are we currently limited to a single verilog file in this mode?

vincent-mirian-google (approved):

flows/yosys/synth.tcl
nanavati: Yes. yosys_synth_library_step will be a different rule (because it will need to juggle a filelist). That annoys me a bit, but I comfort myself with knowing that we don't need to keep the parallel paths downstream from here.

flows/yosys/build_defs.bzl
nanavati: I went back and forth on that because I (mistakenly) thought there wasn't an easy way to undo the quiet flag once it is baked into the wrapper script. While Yosys's -v <level> isn't quite the inverse -v is with other tools, I think it's good enough for people to turn up logging when running the wrapped tool (without having to change things inside the wrapper).

Anyway, I've made the change to quieter by default.

PR #92, opened 2022-02-11
Fixes observed nondeterminism in synthesis

QuantamHD:
The dff flag passed to abc appears to cause non-determinism in the generated netlist.


mithro:
I think this needs an associated github issue tracking what would need to change for `-dff` to be readded?

mithro:
If possible it would be great to have a test which verifies everything now stays deterministic. Don't know how hard that would be?

PR #99, opened 2022-03-27
Adds support for the OSX-default Clang.

RobSpringer:
Adds a Bazel `select` for Darwin to specify the OSX-default definition
of the "--export-dynamic/-export_dynamic" linker flag.

Encountered while building iverilog on an M1 Mac. Tested in that environment, but not in Linux and/or on x86.


QuantamHD:
Thanks for the change. I'll let the CI run then merge

PR #105, opened 2022-06-03
Add generated `_synth_output` to the runfiles of `synthesize_rtl`.

grotival:
Add generated `_synth_output` to the runfiles of `synthesize_rtl`.


QuantamHD:
When you build this rule does the log still appear as built file. It's generally useful for debugging

grotival:
yes, all good! it was in the output files, just not in the rufiles

PR #110, opened 2022-08-12
clang and gcc

renau:
The patch disables openmp package and uses the default in the system. Otherwise, it requires the same clang version as the openmp (which is a problem to compile with gcc).

The other part of the patch is to enable c++17 in openroad (they use this in their builds). It does not build in the latest gcc/clang otherwise.

The patch allows gcc and clang to work (I tried gcc 11 and clang 13, but others should work too). To compile with clang:

```
CXX=clang++ CC=clang bazel build ....
```

I create a small repo to test this (it points to renau/bazel_rules_hdl not the hdl/bazel_rules_hdl)

### DIFF
--- a/dependency_support/org_theopenroadproject/bundled.BUILD.bazel
+++ b/dependency_support/org_theopenroadproject/bundled.BUILD.bazel
@@ -115,8 +115,8 @@ cc_binary(
         "-Wcast-qual",
         "-Wredundant-decls",
         "-Wformat-security",
+        "-fopenmp",
         "-std=c++17",
-        "-fopenmp"
     ],
     features = ["-use_header_modules"],
     visibility = ["//visibility:public"],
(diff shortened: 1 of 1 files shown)


QuantamHD:
So there's a couple of things going on here. By default all code should be compiling with c++17 https://github.com/hdl/bazel_rules_hdl/blob/main/.bazelrc#L15, perhaps the run commands need to added to the .bazelrc

The CI runs on a hermetic version of clang so removing the openmp dependency will make that toolchain fail. It would at least need a select based on toolchain.

renau:
OK, I think that I know the issue.
My testing repo does not have the .bazelrc. This means that I do not have an hermetic llvm (I let to use the local in my machine. This allows me to try gcc or clang).

let me try with the bazelrc

PR #116, opened 2022-10-12
Fixes redacted version info in bison and m4

QuantamHD:
This was caused by a weird configuration for config.h in the gnulib package. I added configurability to the version information so that it gives downstream packages the option to override the redacted option.


QuantamHD:
@stridge-cruxml This should fix the bison redacted issue you were running into.

stridge-cruxml:
Thanks for the fix!

PR #120, opened 2022-11-14
Port xsim_test from https://github.com/CruxML/rules_xilinx.

stridge-cruxml:
xsim_test is defined in vivado/defs.bzl. This can run and generate a waveform. Manually tested waveform generation, passing and failing conditions.


vivado/defs.bzl
```diff
@@ -618,3 +646,97 @@ def vivado_flow(name, module, module_top, part_number, xilinx_env, tags = []):
...
+    outputs.append(vivado_journal)
+
+    log_runfiles = ctx.runfiles(files = [vivado_log])
+
+    error_parser = "if grep -q Error {}; then\n".format(vivado_log.short_path)
+    error_parser += "cat {}\n".format(vivado_log.short_path)
+    error_parser += "exit 64\n"
+    error_parser += "fi"
+
+    ctx.actions.write(
```
QuantamHD, line 694: We should move this error checker to a validation action.

https://bazel.build/extending/rules#validation_actions

stridge-cruxml: I hadn't seen those thanks!

I am a bit unsure about how to make a test with a validating action though. It seems like it's meant as a consistency check on building a blob or some other artifact. For a test rule, it is required to return an executable to evaluate if the test passed or failed.

To use a validating action in a test rule I would need to somehow make it generate an executable file to return in the default info. If this is possible, could you point me to an example or provide a code snippet?

QuantamHD: That might be an issue. I'll mark this as resolved for now

PR #121, opened 2022-12-09
Updating bazel build rules to support detailed

QuantamHD:
Updating bazel build rules to support detailed


mithro (commented):
CI is failing?

PR #124, opened 2022-12-13
Add vitis rules for HLS.

stridge-cruxml:


stridge-cruxml:
Not sure what this message means: `[_USERNAME -> $(commit.author.login)]: generic::invalid_argument: failed ParseJSONPath: author is not found`.

Also gentle ping :)

QuantamHD:
I also added you as a maintainer so you can merge things now.

vitis/v2020_1/ap_common.h
```diff
@@ -0,0 +1,764 @@
...
+#-  liability) for any loss or damage of any kind or nature
+#-  related to, arising under or in connection with these
+#-  materials, including for any direct, or any indirect,
+#-  special, incidental, or consequential loss or damage
+#-  (including loss of data, profits, goodwill, or any type of
+#-  loss or damage suffered as a result of any action brought
+#-  by a third party) even if such damage or loss was
+#-  reasonably foreseeable or Xilinx had been advised of the
+#-  possibility of the same.
+#-
```
QuantamHD, line 31: We can't add non-open source code to the repo.

QuantamHD:
This change is abnormally large, and might be causing some issues with Google Cloud build. I'll need to debug more.

vitis/v2020_1/ap_common.h
stridge-cruxml: Ok, this is an issue. Ill have to think about how I can point to a copy in vivado somehow.

This library is installed with vitis hls in a path like: `Vitis_HLS/2021.2/include/` and all cc includes for hls come from there.
Maybe some .bazelrc setting or something.

stridge-cruxml:
> This change is abnormally large, and might be causing some issues with Google Cloud build. I'll need to debug more.

After i remove the cpp headers, it will be small

vitis/v2020_1/ap_common.h
QuantamHD: We could also explore adding it in a toolchain configuration. https://bazel.build/extending/toolchains

PR #142, opened 2023-02-05
Rules to run opensta to get longest path.

tcal-x:
Rules to run opensta to get longest path.

Uses a default "sta.tcl" script that can be overridden.


tcal-x:
This was for testing the copybara config (it did catch a couple of issues that I fixed).  I think we then run copybara again "for real" after my CL is merged inside Google.

PR #146, opened 2023-02-07
Rules to run opensta to get longest path.

tcal-x:
Rules to run opensta to get longest path.

Uses a default "sta.tcl" script that can be overridden.

### DIFF
--- a/static_timing/build_defs.bzl
+++ b/static_timing/build_defs.bzl
@@ -35,7 +35,7 @@ def _run_opensta_impl(ctx):
     opensta_runfiles_dir = ctx.executable._opensta.path + ".runfiles"
 
     sta_tcl = ctx.file.sta_tcl
-    sta_log = ctx.actions.declare_file("sta.log")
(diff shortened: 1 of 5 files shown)


static_timing/build_defs.bzl
```diff
@@ -0,0 +1,98 @@
...
+                direct = [sta_log],
+                transitive = [],
+            ),
+        ),
+    ]
+
+run_opensta_attrs = {
+    "synth_target": attr.label(
+        doc = "The synth target to benchmark.",
+        providers = [SynthesisInfo, DefaultInfo],
```
QuantamHD, line 71: It doesn't look like you're grabbing anything from DefaultInfo so it's best to just leave this as SynthesisInfo

QuantamHD (approved):

static_timing/build_defs.bzl
tcal-x: Agreed, I confirmed it can be removed, and also removed it from the CL.

PR #148, opened 2023-02-10
Don't hardcode assumed path to /bin/bash, use /usr/bin/env bash instead.

hzeller:
In a pure environment (such as nixos), the hard-coded path /bin/bash might not exist. The recommended way in the Unix environment is to use /usr/bin/env

Signed-off-by: Henner Zeller <h.zeller@acm.org>


hzeller:
Noticed this as it was a dependency of XLS.

QuantamHD (approved):

PR #155, opened 2023-04-04
Move commands in StaApp into sta:: namespace to match behavior of external builds.

tcal-x:
Move commands in StaApp into sta:: namespace to match behavior of external builds.


tcal-x:
This PR includes an unrelated update to `synthesis/synth.tcl`.

QuantamHD (approved):

PR #156, opened 2023-04-11
Update so that sky130_fd_sc_hd:sky130_fd_sc_hd returns the default corner Liberty file in its DefaultInfo.

tcal-x:
Update so that sky130_fd_sc_hd:sky130_fd_sc_hd returns the default corner Liberty file in its DefaultInfo.


tcal-x:
@QuantamHD  please take a look.

PR #158, opened 2023-05-04
Add ip block creation rule for vivado.

stridge-cruxml:
Vivado allows code to be packaged into 'IP'. There are advantages to packaging like this such as:

- Use within vivado block design (a GUI to connect these blocks together).
- Encryption - Source files can be encrypted with vivado holding decryption keys (implemented with this PR).
- Payment - A IP block can be distributed requiring a license within vivado.
- Logical organization/separation of code

Internally, we have used this to successfully build an encrypted IP block and also build a bitstream using that block. See [here](https://github.com/hdl/bazel_rules_hdl/blob/2896461b078a00ec7f1db4d0e4f119dcf7cc931e/vivado/tests/BUILD#L116) in the tests for an example creating an encypted IP block.


QuantamHD:
LGTM!

QuantamHD (approved):

PR #162, opened 2023-05-23
org_llvm_openmp: Add omp-tools.h to bundled.BUILD.bazel

lpawelcz:
This PR fixes `org_llvm_openmp` build failures I encountered while trying to run target based on rule `place_and_route`. 

`BUILD` file for `openmp` didn't specify `omp-tools.h` as its dependency. This header is present in the codebase as `omp-tools.h.var` and it is renamed to required `omp-tools.h` by the [cmake flow](https://github.com/llvm/llvm-project/blob/main/openmp/runtime/src/CMakeLists.txt#L17). In order to fix this in bazel rules I added `genrule` for copying the file and specified the missing dependency.

### DIFF
--- a/dependency_support/org_llvm_openmp/bundled.BUILD.bazel
+++ b/dependency_support/org_llvm_openmp/bundled.BUILD.bazel
@@ -14,7 +14,10 @@
 
 cc_library(
     name = "openmp",
-    hdrs = ["runtime/src/include/omp.h"],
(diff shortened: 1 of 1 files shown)


QuantamHD:
Thanks for contributing!

QuantamHD:
No changes needed thanks again..

PR #164, opened 2023-05-24
dependency_support: boost: bump boost to v1.82

lpawelcz:
This PR is the second one in the PR chain that consists of:
* [1/3] https://github.com/hdl/bazel_rules_hdl/pull/167
* [2/3] https://github.com/hdl/bazel_rules_hdl/pull/164
* [3/3] https://github.com/hdl/bazel_rules_hdl/pull/165

Each of the following PRs depend on previous ones which means that the correct merging order is: 
https://github.com/hdl/bazel_rules_hdl/pull/167 -> https://github.com/hdl/bazel_rules_hdl/pull/164 -> https://github.com/hdl/bazel_rules_hdl/pull/165
I will rebase the PRs one at the time as they are merged in order to avoid conflicts.

This PR bumps the build rules for boost effectively bumping boost dependency to v1.82.
Due to rename (BUILD.boost -> boost.BUILD) and small changes in the renamed file also the patches had to be updated.
Additionally, `boost_library` for `boost/python` adds to `cc_library` sources all files under `libs/python/src` which includes `fabscript` which is not a C++ source and that is causing build failures. I had to further modify `add_python.patch` to explicitly remove this file from the build configuration.

Bumping boost was required to make `openroad` target working as I kept getting build errors:
```
ERROR: /home/asdf/.cache/bazel/asdf/970c5c2433bb6038ab152477a024c421/external/boost/BUILD.bazel:1750:14: Compiling libs/thread/src/pthread/thread.cpp failed: (Exit 1): gcc failed: error executing command /usr/bin/gcc -U_FORTIFY_SOURCE -fstack-protector -Wall -Wunused-but-set-parameter -Wno-free-nonheap-object -fno-omit-frame-pointer -g0 -O2 '-D_FORTIFY_SOURCE=1' -DNDEBUG -ffunction-sections ... (remaining 312 arguments skipped)

Use --sandbox_debug to see verbose messages from the sandbox
In file included from /usr/include/pthread.h:33,
                 from /usr/include/x86_64-linux-gnu/c++/11/bits/gthr-default.h:35,
                 from /usr/include/x86_64-linux-gnu/c++/11/bits/gthr.h:148,
                 from /usr/include/c++/11/ext/atomicity.h:35,
                 from /usr/include/c++/11/bits/basic_string.h:39,
                 from /usr/include/c++/11/string:55,
                 from external/boost/boost/thread/exceptions.hpp:20,
                 from external/boost/boost/thread/pthread/thread_data.hpp:10,
                 from external/boost/boost/thread/thread_only.hpp:17,
                 from external/boost/libs/thread/src/pthread/thread.cpp:11:
external/boost/boost/thread/pthread/thread_data.hpp:60:5: error: missing binary operator before token "("
   60 | #if PTHREAD_STACK_MIN > 0
      |     ^~~~~~~~~~~~~~~~~
Target @org_theopenroadproject//:openroad failed to build
```

Another required change was bumping `spdlog`. After bumping boost I encountered spdlog build errors in my local test setup that were caused by `MINSIGSTKSZ` no longer being a constant (as described in https://github.com/catchorg/Catch2/issues/2421 and https://github.com/gabime/spdlog/issues/2058)

### DIFF
--- a/dependency_support/boost/add_python.patch
+++ b/dependency_support/boost/add_python.patch
@@ -12,12 +12,12 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-diff --git BUILD.boost BUILD.boost
-index e6b28cf..99470c5 100644
---- BUILD.boost
-+++ BUILD.boost
-@@ -2268,3 +2268,41 @@ boost_library(
-         ":utility",
+diff --git boost.BUILD boost.BUILD
+index e5d0b60..4f52d3c 100644
+--- boost.BUILD
++++ boost.BUILD
+@@ -2642,3 +2642,44 @@ boost_library(
+         ":variant2",
      ],
  )
 +
@@ -27,6 +27,9 @@ index e6b28cf..99470c5 100644
 +        "libs/python/src/converter/*.cpp",
 +        "libs/python/src/object/*.cpp",
 +    ]),
++    exclude_src = [
++        "libs/python/src/fabscript",
++    ],
 +    deps = [
 +        ":assert",
 +        ":bind",
--- a/dependency_support/boost/boost.bzl
+++ b/dependency_support/boost/boost.bzl
@@ -21,8 +21,8 @@ def boost():
     maybe(
         git_repository,
         name = "com_github_nelhage_rules_boost",
-        # This equivalent to boost 1.78
-        commit = "c8b9b4a75c4301778d2e256b8d72ce47a6c9a1a4",
+        # This equivalent to boost 1.82
+        commit = "1217caae292dc9f14e8109777ba43c988cf89c5b",
         remote = "https://github.com/nelhage/rules_boost",
         shallow_since = "1640124117 -0800",
         patches = [
--- a/dependency_support/boost/libbacktrace.patch
+++ b/dependency_support/boost/libbacktrace.patch
@@ -1,9 +1,9 @@
-diff --git BUILD.boost BUILD.boost
-index e23893b..1e70e11 100644
---- BUILD.boost
-+++ BUILD.boost
-@@ -1874,6 +1874,7 @@ boost_library(
-         ":lexical_cast",
+diff --git boost.BUILD boost.BUILD
+index e5d0b60..bade47f 100644
+--- boost.BUILD
++++ boost.BUILD
+@@ -1927,6 +1927,7 @@ boost_library(
+         ":predef",
          ":static_assert",
          ":type_traits",
 +        "@com_github_libbacktrace//:libbacktrace",
--- a/dependency_support/com_github_gabime_spdlog/com_github_gabime_spdlog.bzl
+++ b/dependency_support/com_github_gabime_spdlog/com_github_gabime_spdlog.bzl
@@ -22,9 +22,9 @@ def com_github_gabime_spdlog():
         http_archive,
         name = "com_github_gabime_spdlog",
         urls = [
-            "https://github.com/gabime/spdlog/archive/refs/tags/v1.9.2.tar.gz",
+            "https://github.com/gabime/spdlog/archive/refs/tags/v1.11.0.tar.gz",
         ],
-        strip_prefix = "spdlog-1.9.2",
-        sha256 = "6fff9215f5cb81760be4cc16d033526d1080427d236e86d70bb02994f85e3d38",
+        strip_prefix = "spdlog-1.11.0",
+        sha256 = "ca5cae8d6cac15dae0ec63b21d6ad3530070650f68076f3a4a862ca293a858bb",
         build_file = Label("@rules_hdl//dependency_support/com_github_gabime_spdlog:bundled.BUILD.bazel"),
     )


QuantamHD:
Looks like there's something off with the Boost rules

```
Step #1: ERROR: Traceback (most recent call last):
Step #1: 	File "/builder/home/.cache/bazel/_bazel_root/eab0d61a99b6696edb3d2aff87b585e8/external/boost/BUILD.bazel", line 13, column 30, in <toplevel>
Step #1: 		_repo_dir = repository_name().removeprefix("@")
Step #1: Error: 'string' value has no field or method 'removeprefix'

```

lpawelcz:
Oh yes, that is indeed bazel version issue. It looks like the rules for building bazel starting from https://github.com/nelhage/rules_boost/commit/5729d34dcf595874f32b9f1aa1134db65fe78fda#diff-fa9d836e1eb513176a2b1b9e121bcde61bfed880a680a8690ef84fc65b92f0bdR13 will work only with bazel releases >= v5.1.0 after https://github.com/bazelbuild/starlark/issues/185 was solved.
I wanted to check if we can solve this the other way around - by downgrading boost build rules just bellow the linked commit (so that we don't need the `removeprefix()` method) and staying at bazel v4.2.1 but I ran into other errors:
```
ERROR: Analysis of target '//:counter_place_and_route' failed; build aborted: @platforms//os:osx is not a valid select() condition for @org_bzip_bzip2//:bz2lib.
To inspect the select(), run: bazel query --output=build @org_bzip_bzip2//:bz2lib.
For more help, see https://docs.bazel.build/be/functions.html#select.
```

This error does not occur on bazel v5.1.0.
How hard would it be to bump bazel version here in `bazel_rules_hdl`? Is it only the matter of changing the container image from https://github.com/hdl/bazel_rules_hdl/blob/af96df4384cc237608be1ac249db55602080dc88/cloudbuild.yaml#L36 to `gcr.io/bazel-public/bazel:latest` (or better, to pin the version to v5.1.0 or newer)?

proppy:
> How hard would it be to bump bazel version here in bazel_rules_hdl? Is it only the matter of changing the container image from

I think we would also need to document the new required version here: https://github.com/hdl/bazel_rules_hdl/blob/main/README.md?plain=1#L4

proppy:
different failure here:
```
Step #1: W: chmod 0700 of directory /var/lib/apt/lists/partial failed - SetupAPTPartialDirectory (1: Operation not permitted)
Step #1: E: Could not open lock file /var/lib/apt/lists/lock - open (13: Permission denied)
Step #1: E: Unable to lock directory /var/lib/apt/lists/
```

cloudbuild.yaml
```diff
@@ -42,12 +42,6 @@ steps:
     apt-get update
     apt-get install apt-transport-https ca-certificates -y
```
proppy, line 43: we can remove those as well, as we're not fetching bazel anymore.

lpawelcz:
I don't have the access to build logs from Google Cloud Build but it looks like it fails on early stage of the build. It might be caused by changing the container image. I will look into this.

cloudbuild.yaml
lpawelcz: I'm not sure if we can do that but I'll try that out. Those lines were added on OpenROAD bump 2 years ago: https://github.com/hdl/bazel_rules_hdl/pull/56

proppy: ah interesting I though they were added because of the `wget`, I think we should be able to assume that the official bazel image has everything needed to resolve `http_archive` from the `WORKSPACE`.

proppy:
Now the job fails with:
```
Step #1: bash: line 1: build_invocation_id.txt: Permission denied
```

cloudbuild.yaml
```diff
@@ -33,21 +33,12 @@ steps:
...
-    apt-get update
-    apt-get install apt-transport-https ca-certificates -y
-    update-ca-certificates
-    # This is kind of a hack to upgrade bazel to 4.2.1. When a new Bazel Docker container is released https://github.com/bazelbuild/continuous-integration/issues/1060
-    # Remove this, and just bump the docker container.
-    wget https://github.com/bazelbuild/bazel/releases/download/4.2.1/bazel_nojdk-4.2.1-linux-x86_64
-    chmod +x bazel_nojdk-4.2.1-linux-x86_64
-    mv bazel_nojdk-4.2.1-linux-x86_64 /usr/local/lib/bazel/bin/bazel 
-    # End Bazel upgrade hack.
     python3 tools/generate_uuid.py > build_invocation_id.txt
```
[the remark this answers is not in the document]
proppy: actually since it is only used in variable substitution below, it's probably safe to just story to a local env var.

lpawelcz:
I messed up with those env vars:
```
generic::invalid_argument: invalid value for 'build.substitutions': key in the template "INVOCATION_ID" is not a valid built-in substitution
```

cloudbuild.yaml
```diff
@@ -33,35 +33,26 @@ steps:
...
-    # Remove this, and just bump the docker container.
-    wget https://github.com/bazelbuild/bazel/releases/download/4.2.1/bazel_nojdk-4.2.1-linux-x86_64
-    chmod +x bazel_nojdk-4.2.1-linux-x86_64
-    mv bazel_nojdk-4.2.1-linux-x86_64 /usr/local/lib/bazel/bin/bazel 
-    # End Bazel upgrade hack.
-    python3 tools/generate_uuid.py > build_invocation_id.txt
-    python3 tools/generate_uuid.py > test_invocation_id.txt
+    export BUILD_INVOCATION_ID=$(python3 tools/generate_uuid.py)
+    export TEST_INVOCATION_ID=$(python3 tools/generate_uuid.py)
     echo STATUS: Reporting logs with invocation id $(cat invocation_id.txt) to GitHub.
```
lpawelcz, line 44: `invocation_id.txt` probably does not exist. Not sure why it isn't failing on that

lpawelcz:
> still:
> 
> ```
> generic::invalid_argument: invalid value for 'build.substitutions': key in the template "BUILD_INVOCATION_ID" is not a valid built-in substitution
> ```

yes, I will just go with writing files to `/tmp`

proppy:
No need to use `/tmp`, I think env variables require to be specified as `$$` in cloud build configuration, see:
https://cloud.google.com/build/docs/configuring-builds/substitute-variable-values for more details.

lpawelcz:
Thank you @proppy , `$$` works but now it fails inside GCB and I don't have the access to the logs so I cannot see what went wrong

proppy:
> it fails inside GCB

```
cat: invocation_id.txt: No such file or directory
STATUS: Reporting logs with invocation id to GitHub.
FATAL: mkdir('/builder/home/.cache/bazel/_bazel_ubuntu'): (error: 13): Permission denied
```

lpawelcz:
It probably still throws:
```
FATAL: mkdir('/builder/home/.cache/bazel/_bazel_ubuntu'): (error: 13): Permission denied
```

proppy:
Yes:
```
FATAL: mkdir('/builder/home/.cache/bazel/_bazel_ubuntu'): (error: 13): Permission denied
```

lpawelcz:
The issue is that currently used docker image (`l.gcr.io/google/bazel:3.5.0`) has default user set to `root`, while `gcr.io/bazel-public/bazel:5.1.0` has a regular user named `ubuntu`.
It looks like Google Cloud Build default workdir is `/builder` and that creates permission problems for `ubuntu` user in newer image.
In https://github.com/hdl/bazel_rules_hdl/pull/164/commits/754a29fe80298a9df2c99d0b118c6850e8f93257 I set the `dir` attribute to `test` which, according to [the docs](https://cloud.google.com/build/docs/build-config-file-schema#dir), should set the workdir path to `/workspace/test`
Error log from testing this change (https://github.com/hdl/bazel_rules_hdl/pull/164/checks?check_run_id=14075987268) would be quite interesting to see. I would expect error:
```
FATAL: mkdir('/workspace/test/home/.cache/bazel/_bazel_ubuntu'): (error: 13): Permission denied
```

I managed to find another repository with bazel docker images: https://console.cloud.google.com/artifacts/docker/cloud-builders/us/gcr.io/bazel which is based on https://github.com/GoogleCloudPlatform/cloud-builders/tree/master/bazel.
Those images have default user set to `root` but they are not version-tagged. I tested the `latest` tag and it has `bazel v5.4.0` .
With https://github.com/hdl/bazel_rules_hdl/pull/164/commits/f817bbe2361c7a80257f97836188c1d22947ef9d I was able to finally get some bazel output logs in CI run: https://app.buildbuddy.io/invocation/25d8b96a-0539-11ee-a19f-0242c0a80a02. I actually faced those errors locally so it looks like with this docker image we finally got working CI setup

lpawelcz:
Ok, thank you @proppy. It looks like `dir` doesn't change the `/builder` path. This setting must be responsible for the working directory where the repos are cloned, etc. I searched for any mention of `/builder` in Google Cloud Build docs and bazel_rules_hdl but I wasn't successful.

For the time being I will set the docker image to the latest one from https://console.cloud.google.com/artifacts/docker/cloud-builders/us/gcr.io/bazel and I will work on solving issues during the actual bazel_rules_hdl test.

proppy:
alternative there is also https://edu.chainguard.dev/chainguard/chainguard-images/reference/bazel/overview/

proppy (approved):
should we also document the other changes in the PR description (spdlog and libbacktrace deps)?

proppy:
@QuantamHD can you also review (I don't have merging right in this project)

lpawelcz:
Docker images from `cloud-marketplace-containers` are outdated. The last image uploaded [there](https://console.cloud.google.com/gcr/images/cloud-marketplace-containers/GLOBAL/google/bazel) is from 3 sept 2020.
Images from chainguard have `non-root` default user which will again cause problems with permissions.
I'm writing detailed updates for each of the PRs: 
* [1/3] https://github.com/hdl/bazel_rules_hdl/pull/167
* [2/3] https://github.com/hdl/bazel_rules_hdl/pull/164
* [3/3] https://github.com/hdl/bazel_rules_hdl/pull/165

And I will post those shortly

lpawelcz:
I have updated the description. @proppy, @QuantamHD please take a look at this PR and also at https://github.com/hdl/bazel_rules_hdl/pull/167 where I've changed the docker image in order to run CI with bazel v5.4.0

proppy (approved):

proppy:
LGTM, but @QuantamHD needs to hit the merge button :)

QuantamHD (approved):

QuantamHD:
Thanks for the contribution! Looks like there are merge conflicts to resolve

Issue #54, opened 2021-10-15
Support GDSII Output for Place and Route Rules


QuantamHD:
We should use the magic instance we already built to stream the output DEF files into GDSII

QuantamHD:
Thanks for working on this! Could we start with creating it as a separate rule. Inside Google we have a different dependency manager that may or may not play well with the Klayout python module. I'd like to get that working without blocking a merge in this repo.

Once we have it working it shouldn't be a big lift to get it merged together. Thanks again! :)

Issue #20, opened 2020-12-05
Add KLayout


mithro:
Needed for things like DRC checks and GDS generation.

umarcor:
Ref: [hdl.github.io/awesome/items/klayout](https://hdl.github.io/awesome/items/klayout/)

lpawelcz:
Hi, I would like to work on this one. However, It looks like we could use just the [KLayout python module](https://www.klayout.org/klayout-pypi/). This way we don't need to include to the workspace a big dependency which is Qt.

QuantamHD:
Cool. If the python module works we can use that. Thanks for working on it.

PR #169, opened 2023-06-16
Add rule for writing GDS files

lpawelcz:
Closes https://github.com/hdl/bazel_rules_hdl/issues/20, closes https://github.com/hdl/bazel_rules_hdl/issues/54.

This PR adds `gds_write` rule which allows writing GDS files. The rule have a set of inputs and dependencies which are:
* **Implemented RTL** - the result of `place_and_route` target,
* **KLayout** - delivered as Python module
* **Python script** - for writing GDS files with KLayout module
* **KLayout technology file**  - XML file generated from KLayout GUI version, based on data from PDKs.
* **KLayout LEF file** - technology LEF file of the PDK that includes all relevant information regarding metal layers, vias, and spacing requirements
* **Additional LEF files** -  List of hardened macro LEF view files.
* **Additional GDS files** - List of platform GDS files
* **Fill config** - JSON file with configuration for metal fill

Apart from the rules itself, PR:
* adds KLayout python module into `pip_requirements.txt`
* imports and modifies [def2stream.py](https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/blob/99e8d3e9b399c7703fab26bef2b5770d39a2f724/flow/util/def2stream.py) script from OpenROAD-flow-scripts
* modifies `place_and_route` rule to also return `SynthesisInfo` provider which is convenient for accessing top module name for the GDS write script.
* modifies `StandardCellInfo` provider to also store GDS files from PDKs
* adds `density_fill` stage after `detailed_routing`


lpawelcz:
Hi @QuantamHD and @proppy, please take a look at this PR

gds_write/build_defs.bzl
```diff
@@ -0,0 +1,127 @@
...
+        outputs = [
+            klayout_final_lyt
+        ],
+        inputs = depset(
+            [ctx.file.klayout_lyt, klayout_final_lef] +
+            final_lefs
+        ),
+        # Replace existing entry with reference to correct KLayout LEF file
+        # Save the results in klayout_final.lyt in bazel build directory
+        command =
```
proppy, line 60: curious if that could be done if def2stream instead?

gds_write/build_defs.bzl
```diff
@@ -0,0 +1,127 @@
...
+def _gds_write_impl(ctx):
+    final_gds = ctx.actions.declare_file("{}.gds".format(ctx.attr.name))
+    klayout_final_lyt = ctx.actions.declare_file("{}_klayout.lyt".format(ctx.attr.name))
+    klayout_final_lef = ctx.actions.declare_file("{}_klayout.lef".format(ctx.attr.name))
+    # Declare a list of additional LEF files to be copied into bazel build directory
+    final_lefs = []
+    for lef in ctx.files.additional_lef:
+        final_lefs.append(ctx.actions.declare_file("{}_{}.lef".format(ctx.attr.name, lef.basename.split('.', 1)[0])))
+
+    # Copy additional LEF files to bazel build directory
```
proppy, line 29: curious why we need to copy the lef file? can't we just use them from their current location?

gds_write/build_defs.bzl
```diff
@@ -0,0 +1,127 @@
...
+        command =
+            "sed " +
+            "\"s/<lef-files>.*<\\/lef-files>/<lef-files>{}<\\/lef-files>/g\" ".format(klayout_final_lef.basename) +
+            "{} ".format(ctx.file.klayout_lyt.path) +
+            "> {}".format(klayout_final_lyt.path),
+        mnemonic = "FixPathInKlayoutTechFile"
+    )
+
+    # PYTHONPATH environment variable with python script imports
+    pythonpath = ""
```
proppy, line 69: is this necessary to resolve the new `klayout` requirements?

gds_write/build_defs.bzl
```diff
@@ -0,0 +1,127 @@
...
+    for imp in ctx.attr._gds_write[PyInfo].imports.to_list():
+        pythonpath += ":{}".format(imp)
+
+    # Write GDS generation shell script
+    additional_gds_paths = " ".join([file.path for file in ctx.files.additional_gds])
+    write_gds_script = ctx.actions.declare_file("write_gds.sh")
+    write_gds_cmd = "PYTHONPATH=$PYTHONPATH:{} python {}".format(pythonpath, ctx.executable._gds_write.path) + \
```
proppy, line 76: curious why we need an intermediate script rather than `run_shell`?

gds_write/def2stream.py
```diff
@@ -0,0 +1,208 @@
+#!/usr/bin/env python3
+
+import klayout.db as db
```
proppy, line 3: need that you're able to use the klayout python module rather than the full executable!

gds_write/def2stream.py
```diff
@@ -0,0 +1,208 @@
...
+import os
+import argparse
+
+parser = argparse.ArgumentParser(description="GDS writer script")
+parser.add_argument("-d", "--design-name",    required=True,  help="Name of the top module in the design")
+parser.add_argument("-i", "--input-def",      required=True,  help="Path to def file of the implemented design after detailed routing step")
+parser.add_argument("-t", "--tech-file",      required=True,  help="Path to KLayout technology file, a mapping from LEF/DEF to GDS.")
+parser.add_argument("-o", "--out",            required=True,  help="Path to output GDS file")
+parser.add_argument("-g", "--additional-gds", required=False, help="Paths to additional input GDS files", default="", nargs="*")
+parser.add_argument("-f", "--fill-config",    required=False, help="Path to JSON rule file for metal fill during chip finishing", default="")
```
proppy, line 17: curious why we don't use openroad for that?
https://openroad.readthedocs.io/en/latest/main/src/fin/README.html

gds_write/def2stream.py
```diff
@@ -0,0 +1,208 @@
...
+parser.add_argument("-s", "--seal",           required=False, help="Path to seal GDS/OAS file", default="")
```
proppy, line 18: what is `seal` for?

gds_write/def2stream.py
```diff
@@ -0,0 +1,208 @@
...
+   \(\ (?P<xlo>\d+)\ (?P<ylo>\d+)\ \)\   # rect lower-left pt
+  \(\ (?P<xhi>\d+)\ (?P<yhi>\d+)\ \)\ ; # rect upper-right pt
+  ''',
+                      re.VERBOSE)
+
+def read_fills(top):
+  if args.fill_config == '':
+    print('WARNING: no fill config file specified')
+    return
+  # KLayout doesn't support FILL in DEF so we have to side load them :(
```
proppy, line 83: is there an issue on the klayout repo already opened for that?

gds_write/def2stream.py
```diff
@@ -0,0 +1,208 @@
...
+# Load def file
+main_layout = db.Layout()
+print("[INFO] Reporting cells prior to loading DEF ...")
+for i in main_layout.each_cell():
+  print("[INFO] '{0}'".format(i.name))
+
+print("[INFO] Reading DEF ...")
+main_layout.read(args.input_def, layoutOptions)
+
+#print("[INFO] Reporting cells after loading DEF ...")
```
proppy, line 130: leftovers?

gds_write/def2stream.py
```diff
@@ -0,0 +1,208 @@
...
+
+# Clear cells
+top_cell_index = main_layout.cell(args.design_name).cell_index()
+
+# remove orphan cell BUT preserve cell with VIA_
+#  - KLayout is prepending VIA_ when reading DEF that instantiates LEF's via
+print("[INFO] Clearing cells...")
+for i in main_layout.each_cell():
+  if i.cell_index() != top_cell_index:
+    if not i.name.startswith("VIA_"):
```
proppy, line 142: wouldn't the `VIA_` also be child of the top_cell in the DEF?

gds_write/def2stream.py
```diff
@@ -0,0 +1,208 @@
...
+for fil in args.additional_gds:
+  print("\t{0}".format(fil))
+  main_layout.read(fil)
+
+# Copy the top level only to a new layout
+print("[INFO] Copying toplevel cell '{0}'".format(args.design_name))
+top_only_layout = db.Layout()
+top_only_layout.dbu = main_layout.dbu
+top = top_only_layout.create_cell(args.design_name)
+top.copy_tree(main_layout.cell(args.design_name))
```
proppy, line 156: if we only copy the top level, did we actually need to clear the orphan cells?

gds_write/def2stream.py
```diff
@@ -0,0 +1,208 @@
...
+
+read_fills(top)
+
+print("[INFO] Checking for missing cell from GDS/OAS...")
+missing_cell = False
+regex = None
+if 'GDS_ALLOW_EMPTY' in os.environ:
```
proppy, line 163: should we use a flag instead?

gds_write/def2stream.py
```diff
@@ -0,0 +1,208 @@
...
+  for cell in top_only_layout.top_cells():
+    if cell != top_cell:
+      print("[INFO] Merging '{0}' as child of '{1}'".format(cell.name, top_cell.name))
+      top.insert(db.CellInstArray(cell.cell_index(), db.Trans()))
+
+# Write out the GDS
+print("[INFO] Writing out GDS/OAS '{0}'".format(args.out))
+top_only_layout.write(args.out)
+
+sys.exit(errors)
```
proppy, line 208: should we use exception instead? so that can exit earlier in the flow?

gds_write/def2stream.py
```diff
@@ -0,0 +1,208 @@
...
+  top_only_layout.read(args.seal)
+
+  for cell in top_only_layout.top_cells():
+    if cell != top_cell:
+      print("[INFO] Merging '{0}' as child of '{1}'".format(cell.name, top_cell.name))
+      top.insert(db.CellInstArray(cell.cell_index(), db.Trans()))
+
+# Write out the GDS
+print("[INFO] Writing out GDS/OAS '{0}'".format(args.out))
+top_only_layout.write(args.out)
```
proppy, line 206: do we actually want to write if we have an error?

gds_write/def2stream.py
```diff
@@ -0,0 +1,208 @@
+#!/usr/bin/env python3
```
proppy, line 1: license header and docstring?

gds_write/BUILD
```diff
@@ -0,0 +1,32 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# GDS write tool package.
+
+load("@rules_hdl_pip_deps//:requirements.bzl", "requirement")
+load("@rules_python//python:defs.bzl", "py_binary")
+
+py_binary(
```
proppy, line 20: should we have some `py_test` too?

synthesis/tests/BUILD
```diff
@@ -42,6 +43,16 @@ place_and_route(
     synthesized_rtl = ":verilog_counter_synth",
 )
 
+gds_write(
+	name = "counter_asic",
+	implemented_rtl = ":counter_place_and_route",
+	klayout_lyt = "klayout.lyt",
+	klayout_lef = "klayout_tech.lef",
+	fill_config = "fill.json",
+	additional_lef = ["sky130_fd_sc_hd_merged.lef"],
```
proppy, line 52: can we get those thru dependencies?

proppy (changes requested):
Sorry for the delay,

Looks very good! I made a first round of comments with nits and questions about the implementation.

gds_write/build_defs.bzl
lpawelcz: Managed to simplify the rule and remove copying files

gds_write/build_defs.bzl
lpawelcz: Turns out we can just set `tools` attribute in `run_shell` action to point to `py_binary` with our GDS generation script and the environment would be correct then.

gds_write/build_defs.bzl
lpawelcz: Managed to change that to single `run_shell`

place_and_route/build_defs.bzl
```diff
@@ -60,6 +60,7 @@ def _place_and_route_impl(ctx):
 
     return [
         DefaultInfo(files = depset(output_files)),
+        SynthesisInfo(top_module = ctx.attr.synthesized_rtl[SynthesisInfo].top_module),
```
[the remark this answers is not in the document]
lpawelcz: It would be helpful if `place_and_route` rule would return SynthesisInfo provider with the name of the `top_module` from the synthesized design. This name is required in GDS write script, see e.g. [here](https://github.com/hdl/bazel_rules_hdl/pull/169/files/1b70039d8ff2dc0e741ba9a020a606b99d9dc023#diff-9a0852b1664f229eada1c6cf7973ba4e8f17edd051dfbe1b4b560bc7c52c3787R135)

gds_write/def2stream.py
lpawelcz: We can do that. I added `density_fill` stage right after `detailed_routing`

gds_write/def2stream.py
lpawelcz: Looks like this issue was solved in https://github.com/KLayout/klayout/issues/673. I modified the script and bazel flow to perform density fill after detailed routing and then write final DEF file which is read during GDS write

gds_write/def2stream.py
```diff
@@ -0,0 +1,208 @@
...
+        continue
+      data = vals[key]
+      if isinstance(data['datatype'], int):
+        data['datatype'] = [data['datatype']] # convert to array
+      data['klayout'] = [main_layout.find_layer(layer, datatype)
+                         for datatype in data['datatype']]
+
+  return cfg
+
+#match a line like:
```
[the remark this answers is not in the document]
lpawelcz: This part of the code was removed with whole density fill support and reimplemented in bazel flow right after detailed routing stage

gds_write/def2stream.py
lpawelcz: Removed leftover comments

gds_write/def2stream.py
lpawelcz: Removed this part of the code as it seems to be unnecessary.

gds_write/def2stream.py
lpawelcz: I guess it is actually unnecessary. I don't see the point of removing the orphans from `main_layout` if we copy one specific cell tree from it and then use this cell through the rest of the flow. Thank you for pointing that out.

gds_write/def2stream.py
lpawelcz: Raising exception is now placed before writing GDS.

gds_write/def2stream.py
lpawelcz: replaced with `raise LookupError()`

gds_write/def2stream.py
```diff
@@ -0,0 +1,208 @@
...
+
+if not missing_cell:
+  print("[INFO] All LEF cells have matching GDS/OAS cells")
+
+print("[INFO] Checking for orphan cell in the final layout...")
+orphan_cell = False
+for i in top_only_layout.each_cell():
+  if i.name != args.design_name and i.parent_cells() == 0:
+    orphan_cell = True
+    print("[ERROR] Found orphan cell '{0}'".format(i.name))
```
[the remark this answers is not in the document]
lpawelcz: For me it looks impossible, especially when we explicitly copy only the tree of the top cell to the new layout and check for orphans in it. I think we can remove this check.

gds_write/def2stream.py
lpawelcz: Replaced with `--gds-allow-empty` flag

synthesis/tests/BUILD
lpawelcz: I'm afraid that if we place all those files under single `dependencies` attribute we would have problems with distinguishing which file is which as they will be stored on a list and we would have to rely on file extensions.

gds_write/BUILD
lpawelcz: [Here](https://github.com/antmicro/bazel_rules_hdl/blob/gds-write/synthesis/tests/BUILD#L47-L53) we test the rule that uses this `py_binary`. If that is not enough I can add `py_test` but I think it would not be much different compared to the linked test. Please let me know what do you think.

gds_write/def2stream.py
lpawelcz: I'm not sure if I fully understand you here. Could you please explain what do you mean? Is there anything to change regarding that import or KLayout Python Module in general?

gds_write/def2stream.py
lpawelcz: Added license and short docstring

lpawelcz:
@proppy I responded to all of your comments. Please take another look at this PR

gds_write/build_defs.bzl
```diff
@@ -0,0 +1,74 @@
...
+
+gds_write = rule(
+    implementation = _gds_write_impl,
+    attrs = {
+        "implemented_rtl": attr.label(mandatory = True, providers = [OpenRoadInfo, SynthesisInfo]),
+        "klayout_lyt": attr.label(mandatory = True, allow_single_file = True),
+        "input_lef": attr.label_list(allow_files = True),
+        "input_gds": attr.label_list(allow_files = True),
+        "gds_allow_empty": attr.string(),
+        "_gds_write": attr.label(
```
proppy, line 66: do we need this to be an attribute?

gds_write/def2stream.py
```diff
@@ -0,0 +1,118 @@
+#!/usr/bin/env python3
+
+# BSD 3-Clause License
+#
+# Copyright (c) 2018-2023, The Regents of the University of California
```
proppy, line 5: is that the original license from which you created the derivative? There might be some implication since bazel_rules_hdl is Apache 2.0, see: https://github.com/hdl/bazel_rules_hdl/blob/main/LICENSE

@QuantamHD any insights?

gds_write/def2stream.py
```diff
@@ -0,0 +1,118 @@
...
+# IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+# DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE
+# FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+# DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+# SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+# CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+# OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+# OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+"""
```
proppy, line 33: should we refer to the origin of the script?

gds_write/def2stream.py
proppy: s/need/neat/, sorry for the confusion :) I'm glad you're using the python module here.

gds_write/def2stream.py
```diff
@@ -0,0 +1,118 @@
...
+        print("[WARNING] LEF Cell '{0}' ignored. Matches gds_allow_empty.".format(i.name))
+    else:
+        print("[ERROR] LEF Cell '{0}' has no matching GDS/OAS cell."
+              " Cell will be empty.".format(i.name))
+        errors += 1
+
+if not missing_cell:
+  print("[INFO] All LEF cells have matching GDS/OAS cells")
+
+if (errors > 0):
```
proppy, line 113: any reason we don't raise the exception in place?

place_and_route/build_defs.bzl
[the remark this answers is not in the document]
proppy: /cc @QuantamHD

place_and_route/private/detailed_routing.bzl
```diff
@@ -55,12 +55,20 @@ def detailed_routing(ctx, open_road_info):
         if detailed_routing_configs.enable_via_gen:
             detailed_routing_args += " -disable_via_gen "
 
-    open_road_commands = timing_setup_command_struct.commands + [
-        "detailed_route -output_drc {} {}".format(output_drc.path, detailed_routing_args),
-        "write_def {}".format(
-            routed_def.path,
-        ),
-    ]
+    open_road_commands = timing_setup_command_struct.commands
```
proppy, line 58: can you comment about why this change is required?

synthesis/tests/klayout.lyt
```diff
@@ -0,0 +1,158 @@
+<?xml version="1.0" encoding="utf-8"?>
```
proppy, line 1: is there a way we can get this from the pdk deps?

synthesis/tests/BUILD
proppy: see https://github.com/hdl/bazel_rules_hdl/tree/main/dependency_support/com_google_skywater_pdk

proppy (approved):
Nice work, just some nits and questions left.

I'll let @QuantamHD a chance to review before hitting the merge button.

gds_write/build_defs.bzl
lpawelcz: This is used to create implicit dependency on def2stream py_binary. `_gds_write` is a private attribute so it is not possible to override its default value. This way we can also refer to the script through `ctx.executable.<_attr_name>`.

I saw a similar solution in [STA rule](https://github.com/hdl/bazel_rules_hdl/blob/main/static_timing/build_defs.bzl#L78-L82) and [Verilator rule](https://github.com/hdl/bazel_rules_hdl/blob/main/verilator/defs.bzl#L209-L214) and I wanted to follow this pattern here as well.

gds_write/def2stream.py
lpawelcz: This license is taken from [OpenROAD-flow-scripts](https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/blob/master/LICENSE_BUILD_RUN_SCRIPTS) but it was never a part of the [source file](https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/blob/master/flow/util/def2stream.py) that I imported. However, the License File states that it applies to all build and run scripts from OpenROAD-flow-scripts so it applies also to `def2stream.py`.

As far as I know, it is possible to import BSD 3-licensed source files into Apache 2.0 project but it is required to keep the original license header in the imported file.

gds_write/def2stream.py
lpawelcz: Added a note on the origins with link to repo, import date, git hash and link to the source file at mentioned revision.

gds_write/def2stream.py
lpawelcz: We can do that but I wanted to let the script print out all the missing cells first before raising the exception.
Let me know if you would prefer to have this script fail early.

synthesis/tests/klayout.lyt
lpawelcz: In skywater-pdk repositories there are no KLayout files. Commit messages in https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/pull/299 indicate that KLayout files were taken from [RTimothyEdwards/open_pdks](https://github.com/RTimothyEdwards/open_pdks) and those in open_pdks are taken from [mabrains/sky130_klayout_pdk](https://github.com/mabrains/sky130_klayout_pdk).

We could possibly add  `mabrains/sky130_klayout_pdk` here as dependency and write some new rules for fetching this file. Please let me know what do you think about that.

gds_write/build_defs.bzl
QuantamHD: LGTM.

place_and_route/build_defs.bzl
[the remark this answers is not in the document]
QuantamHD: Why not just return the ctx.attr.synthesized_rtl[SynthesisInfo] directly there's really no harm.

synthesis/tests/sky130_fd_sc_hd.gds
QuantamHD, line 1: The GDS and LEF should be available in the sky130 dependency. Ideally, we would add them to the SynthesisInfo's StandardCellInfo provider https://github.com/hdl/bazel_rules_hdl/blob/main/pdk/build_defs.bzl#L18

The raw files are collected by this file:
https://github.com/hdl/bazel_rules_hdl/blob/main/dependency_support/com_google_skywater_pdk/declare_cell_library.bzl

I'll approve without this though since this is just a test, but the idea of the bazel rules is that you only need to specify the PDK files in the synthesis rule, and all other files are transitively available to the subsequent rules. Including info like GDS and LEF

QuantamHD:
Sorry for the long poll in review time. I'm committed to get that down to 24 hours. I will be more responsive going forward.

place_and_route/build_defs.bzl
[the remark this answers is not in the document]
lpawelcz: Done

synthesis/tests/BUILD
lpawelcz: Oh, I see. I was able to use PDK dependencies to pass down LEF and GDS files through `StandardCellInfo` provider. From now on there is no need to specify `input_lef` and `input_gds` attributes. However I've left those attributes - they might be useful in some custom cases.

synthesis/tests/sky130_fd_sc_hd.gds
lpawelcz: Thank you, that was really helpful. With your hints I was able to use LEF and GDS files from sky130 dependency (as described in https://github.com/hdl/bazel_rules_hdl/pull/169#discussion_r1258214673). I've also attempted to modify `StandardCellInfo` for `ASAP7` but I'm not sure if this change should land in this PR

synthesis/tests/klayout.lyt
lpawelcz: Klayout *.lyt file has 3 different sources from which it could be imported to [OpenROAD-flow-scripts](https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/blob/master/flow/platforms/sky130hd/sky130hd.lyt):
* [mabrains/sky130_klayout_pdk](https://github.com/mabrains/sky130_klayout_pdk/blob/main/sky130_tech/tech/sky130/sky130.lyt)
* [efabless/sky130_klayout_pdk](https://github.com/efabless/sky130_klayout_pdk/blob/main/sky130_tech/tech/sky130/sky130.lyt)
* [RTimothyEdwards/open_pdks](https://github.com/RTimothyEdwards/open_pdks/blob/master/sky130/klayout/sky130.lyt)

I tested all of them and compared against *.lyt from `ORFS` and it looks like the most similar is the file from `open_pdks`. There are also some differences in GDS files generated with those KLayout configurations. The ones generated from configurations taken from `ORFS` and `open_pdks` look alright but those generated with configurations from `efabless` and `mabrains` do not have any paths between cells on layers `met2`, `met3`, `met4`

So `open_pdks` looks like a good candidate for fetching those files from dependencies. But there is a catch: open_pdks does not support ASAP7 and doesn't store any files for this technology - ORFS does. If we want to get `fill_config.json` and `klayout.lyt` from dependencies then I think ORFS would be the best repository for fetching this kind of data.

QuantamHD (approved):

QuantamHD:
Thanks!

PR #170, opened 2023-07-04
Add a rule for testing hardware designs using Cocotb framework

rw1nkler:
This PR adds the `cocotb_test` rule which may be used to test hardware designs using Cocotb framework. The created rule uses a simple Python script that filters arguments and passes them to the Cocotb Runner. Most of the arguments of the `cocotb_test` rule are passed directly to the runner. Besides that, a simple test with a counter has been added to test the created rule.

### DIFF
--- a/cocotb/BUILD
+++ b/cocotb/BUILD
@@ -1,0 +1,24 @@
+# Copyright 2023 Antmicro
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+load("@rules_hdl_pip_deps//:requirements.bzl", "requirement")
+
+py_binary(
+    name = "cocotb_wrapper",
+    srcs = ["cocotb_wrapper.py"],
+    python_version = "PY3",
+    srcs_version = "PY3",
+    visibility = ["//visibility:public"],
+    deps = [requirement("cocotb")],
+)
--- a/cocotb/cocotb.bzl
+++ b/cocotb/cocotb.bzl
@@ -1,0 +1,265 @@
+# Copyright 2023 Antmicro
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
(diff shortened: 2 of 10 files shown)


cocotb/cocotb.bzl
```diff
@@ -0,0 +1,267 @@
...
+        transitive = [dep[VerilogInfo].dag for dep in transitive_srcs_list],
+    )
+    verilog_srcs = [
+        verilog_info_struct.srcs
+        for verilog_info_struct in transitive_srcs_depset.to_list()
+    ]
+    verilog_files = [src for sub_tuple in verilog_srcs for src in sub_tuple]
+
+    # from normal files
+    verilog_files += ctx.files.verilog_sources
```
QuantamHD, line 66: This can be slow with a large number of files consider using a depset.

https://bazel.build/rules/lib/builtins/depset

cocotb/BUILD
```diff
@@ -0,0 +1,24 @@
+# Copyright 2023 Google LLC
```
QuantamHD, line 1: Should change the copyright header to the true owner. Either yourself or the company you represent.

cocotb/cocotb.bzl
```diff
@@ -0,0 +1,267 @@
...
+        transitive_files = transitive_files,
+    ).merge_all(
+        [dep.default_runfiles for dep in ctx.attr.deps] +
+        [dep.default_runfiles for dep in ctx.attr.test_module] +
+        [dep.default_runfiles for dep in ctx.attr.sim],
+    )
+
+    # specify pythonpath for the script
+    test_module_paths = _remove_duplicates_from_list([module.dirname for module in ctx.files.test_module])
+    pypath = ":".join([str(p) for p in test_module_paths])
```
QuantamHD, line 151: I'm not a 100% sure on this, but I believe you're supposed to use the Bazel Python "Configuration Fragment" to get the pypath. Could definitely be wrong though.

https://bazel.build/extending/rules#configuration_fragments
https://bazel.build/rules/lib/fragments/bazel_py#python_path

rw1nkler: It seems that the `ctx.fragments.bazel_py.python_path` contains information about the `--python-path` flag which seems to have completely different meaning - the location of the python interpreter (see https://github.com/bazelbuild/bazel/blob/8715e9ae6f5c2fd42607018b3adaad14bf601a1c/src/main/starlark/builtins_bzl/common/python/py_executable.bzl#L217) and may be removed (see https://github.com/bazelbuild/bazel/issues/7901). The variable that I set gives the interpreter additional search paths for module files (see https://docs.python.org/3/using/cmdline.html#envvar-PYTHONPATH)

cocotb/cocotb.bzl
rw1nkler: Let me know if the current changes are what you had in mind

QuantamHD: LGTM.

QuantamHD:
I'll merge once the CI passes, and you unmark this as "Draft"

rw1nkler:
I squashed the review commits and marked the PR as "Ready for review"

QuantamHD (approved):

PR #171, opened 2023-07-10
Improve cocotb_test rule

rw1nkler:
This PR adds the following improvements to the existing `cocotb_test` rule:

* uses `short_path` of files instead of `path` to correctly handle the generated files
* removes the `cocotb_wrapper.py` dependency on the `cocotb` to prevent circular imports and double dependencies when the rule is imported into other repositories. I encountered that problem when I tried to use the `cocotb_bus` library in [the PR to XLS repository](https://github.com/google/xls/pull/1057). The `cocotb_bus` library itself relies on `cocotb`, as a result, I obtained one copy of cocotb from the `bazel_hdl_rules` repo and the second from the `xls` repository.
* adds all Python packages and their dependencies to the `PYTHONPATH` to correctly handle imports from other libraries
* makes the `cocotb_wrapper` attribute public to allow for specifying different entry for other cocotb versions or different build setups

### DIFF
--- a/cocotb/BUILD
+++ b/cocotb/BUILD
@@ -12,13 +12,11 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-load("@rules_hdl_pip_deps//:requirements.bzl", "requirement")
-
 py_binary(
     name = "cocotb_wrapper",
     srcs = ["cocotb_wrapper.py"],
     python_version = "PY3",
     srcs_version = "PY3",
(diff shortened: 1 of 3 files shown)


QuantamHD:
Is this ready to go for merging?

rw1nkler:
> Is this ready to go for merging?

@QuantamHD Yes, it's ready for review and merging

QuantamHD (approved):

PR #173, opened 2023-07-24
pdk: open_road_configuration: Get KLayout technology file and metal fill configuration from internal PDK setup

lpawelcz:
This PR modifies open_road_configuration so that it is now possible for a PDK dependency to define additional two fields:
* metal (density) fill configuration JSON file,
* KLayout technology LYT file.

The last one is required for GDS configuration so this field was marked as mandatory. The files are used by rules: `place_and_route` and `gds_write`. Default values specified for each PDK can be overwritten by attributes `density_config` and `klayout_lyt` of those rules.

It is reasonable to move handling of those files to the PDK dependencies setup because these are PDK-specific and can be reused for multiple designs.

CC @proppy


dependency_support/org_theopenroadproject_asap7/bundled.BUILD.bazel
```diff
@@ -59,6 +59,7 @@ open_road_pdk_configuration(
         "M7": "0.5",
     },
     global_routing_signal_layers = "M2-M7",
+    klayout_tech_file = "@rules_hdl//dependency_support/org_theopenroadproject_asap7:asap7.lyt",
```
proppy, line 62: q: no fill config by default in ASAP7?

proppy (approved):
/cc @QuantamHD

dependency_support/org_theopenroadproject_asap7/bundled.BUILD.bazel
lpawelcz: Looking at the flow examples in OpenROAD-flow-scripts - no, there is no `density_fill` stage in the flow targeting ASAP7 so there is no need for specifying the config file

lpawelcz:
Hi @QuantamHD, please take a look at this PR. Merging this would unblock our work in https://github.com/google/xls/pull/1031

QuantamHD (approved):

lpawelcz:
Thanks a lot!

Issue #172, opened 2023-07-19
Add tests for ASIC flow with ASAP7


lpawelcz:
From what I see, when it comes to ASAP7 support in `bazel_rules_hdl` the tests cover only the synthesis step:
https://github.com/hdl/bazel_rules_hdl/blob/e30c65c618dabffdcdeb45ad66e27fbaa2dc573e/synthesis/tests/BUILD#L71-L78

It would be good to add tests that will run `place_and_route` and `gds_write` flows with the results from `verilog_counter_asap7_synth` as input. This is a blocker for https://github.com/google/xls/issues/996

I started work on adding this but I encountered errors in Clock Tree Synthesis. Looks like there are problems with placing `FILLER` cells, here is the log:
```
[INFO ORD-0030] Using 12 thread(s).
[INFO CTS-0049] Characterization buffer is: BUFx4_ASAP7_75t_R.
[INFO CTS-0039] Number of created patterns = 11880.
[INFO CTS-0084] Compiling LUT.
Min. len    Max. len    Min. cap    Max. cap    Min. slew   Max. slew
2           8           1           34          1           64          
[WARNING CTS-0043] 1584 wires are pure wire and no slew degradation.
TritonCTS forced slew degradation on these wires.
[INFO CTS-0046]     Number of wire segments: 7500.
[INFO CTS-0047]     Number of keys in characterization LUT: 1744.
[INFO CTS-0048]     Actual min input cap: 1.
[INFO CTS-0007] Net "clk" found for clock "core_clock".
[INFO CTS-0010]  Clock net "clk" has 4 sinks.
[INFO CTS-0008] TritonCTS found 1 clock nets.
[INFO CTS-0097] Characterization used 1 buffer(s) types.
[INFO CTS-0027] Generating H-Tree topology for net clk.
[INFO CTS-0028]  Total number of sinks: 4.
[INFO CTS-0090]  Sinks will be clustered based on buffer max cap.
[INFO CTS-0030]  Number of static layers: 0.
[INFO CTS-0020]  Wire segment unit: 20000  dbu (5 um).
[INFO CTS-0023]  Original sink region: [(91411, 144720), (96595, 166320)].
[INFO CTS-0024]  Normalized sink region: [(4.57055, 7.236), (4.82975, 8.316)].
[INFO CTS-0025]     Width:  0.2592.
[INFO CTS-0026]     Height: 1.0800.
[WARNING CTS-0045] Creating fake entries in the LUT.
[...]

PR #176, opened 2023-08-24
Enable designs targeting ASAP7 

lpawelcz:
Fixes https://github.com/hdl/bazel_rules_hdl/issues/172

Requires merging https://github.com/hdl/bazel_rules_hdl/pull/178 in order to fix errors from `buildifier`.

This PR:
* bumps ASAP7 dependency,
* changes the default settings of the PDK dependency to use non-scaled variant,
* updates ASAP7 config files (`pdn_config.pdn`, `rc_script.tcl`, `tracks.tcl`) to match non-scaled variant of the PDK,
* adds test which implements physical layout of counter example targeting ASAP7 technology.

@QuantamHD , @proppy please take a look at this

### DIFF
--- a/dependency_support/org_theopenroadproject_asap7/pdn_config.pdn
+++ b/dependency_support/org_theopenroadproject_asap7/pdn_config.pdn
@@ -28,15 +28,11 @@ set_voltage_domain -name {CORE} -power {VDD} -ground {VSS}
 ####################################
 # standard cell grid
 ####################################
-# ASAP7 in its default configuration is multipled by 
-# 4x looking to fix this upstream.
-set multipler 4.0
-
 define_pdn_grid -name {top} -voltage_domains {CORE}
-add_pdn_stripe -grid {top} -layer {M1} -width {0.072} -pitch {2.16} -offset {0} -followpins
-add_pdn_stripe -grid {top} -layer {M2} -width {0.072} -pitch {2.16} -offset {0} -followpins
-add_pdn_stripe -grid {top} -layer {M5} -width {0.48} -spacing {0.288} -pitch {47.52} -offset {1.188}
-add_pdn_stripe -grid {top} -layer {M6} -width {1.1520} -spacing {0.384} -pitch {48} -offset {2.556}
+add_pdn_stripe -grid {top} -layer {M1} -width {0.018} -pitch {0.54} -offset {0} -followpins
+add_pdn_stripe -grid {top} -layer {M2} -width {0.018} -pitch {0.54} -offset {0} -followpins
+add_pdn_stripe -grid {top} -layer {M5} -width {0.12} -spacing {0.072} -pitch {11.88} -offset {0.300}
+add_pdn_stripe -grid {top} -layer {M6} -width {0.288} -spacing {0.096} -pitch {12} -offset {0.513}
 add_pdn_connect -grid {top} -layers {M1 M2}
 add_pdn_connect -grid {top} -layers {M2 M5}
 add_pdn_connect -grid {top} -layers {M5 M6}
--- a/dependency_support/org_theopenroadproject_asap7/rc_script.tcl
+++ b/dependency_support/org_theopenroadproject_asap7/rc_script.tcl
@@ -12,24 +12,24 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-set_layer_rc -layer M1 -capacitance 6.33956E-02 -resistance 1.3889e-01
-set_layer_rc -layer M2 -capacitance 8.02750E-02 -resistance 2.4222e-02
-set_layer_rc -layer M3 -capacitance 1.23273E-01 -resistance 2.4222e-02
-set_layer_rc -layer M4 -capacitance 1.29902E-01 -resistance 1.6778e-02
-set_layer_rc -layer M5 -capacitance 1.19552E-01 -resistance 1.4677e-02
-set_layer_rc -layer M6 -capacitance 1.23114E-01 -resistance 1.0371e-02
-set_layer_rc -layer M7 -capacitance 1.05408E-01 -resistance 9.6720e-03
+# Liberty units are fF,kOhm
+set_layer_rc -layer M1 -capacitance 1.1368e-01 -resistance 1.3889e-01
+set_layer_rc -layer M2 -capacitance 1.3426e-01 -resistance 2.4222e-02
+set_layer_rc -layer M3 -capacitance 1.2918e-01 -resistance 2.4222e-02
+set_layer_rc -layer M4 -capacitance 1.1396e-01 -resistance 1.6778e-02
+set_layer_rc -layer M5 -capacitance 1.3323e-01 -resistance 1.4677e-02
+set_layer_rc -layer M6 -capacitance 1.1575e-01 -resistance 1.0371e-02
+set_layer_rc -layer M7 -capacitance 1.3293e-01 -resistance 9.6720e-03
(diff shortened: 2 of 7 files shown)


dependency_support/org_theopenroadproject_asap7/pdn_config.pdn
```diff
@@ -15,11 +15,11 @@ set ::ground_nets "VSS"
 ####################################
 # global connections
 ####################################
-add_global_connection -defer_connection -net {VDD} -inst_pattern {.*} -pin_pattern {^VDD$} -power
```
proppy, line 18: is that also required because of the non-scaled variant? if not, can you add a line about this in the PR description?

dependency_support/org_theopenroadproject_asap7/rc_script.tcl
```diff
@@ -12,24 +12,24 @@
...
 
-set_layer_rc -layer M1 -capacitance 6.33956E-02 -resistance 1.3889e-01
-set_layer_rc -layer M2 -capacitance 8.02750E-02 -resistance 2.4222e-02
-set_layer_rc -layer M3 -capacitance 1.23273E-01 -resistance 2.4222e-02
-set_layer_rc -layer M4 -capacitance 1.29902E-01 -resistance 1.6778e-02
-set_layer_rc -layer M5 -capacitance 1.19552E-01 -resistance 1.4677e-02
-set_layer_rc -layer M6 -capacitance 1.23114E-01 -resistance 1.0371e-02
-set_layer_rc -layer M7 -capacitance 1.05408E-01 -resistance 9.6720e-03
+# Liberty units are fF,kOhm
+set_layer_rc -layer M1 -capacitance 1.1368e-01 -resistance 1.3889e-01
```
proppy, line 16: q: is that expected that the resistance doesn't change for the non-scaled version?

dependency_support/org_theopenroadproject_asap7/rc_script.tcl
```diff
@@ -12,24 +12,24 @@
...
-set_layer_rc -via V1 -resistance 1.00E-02
-set_layer_rc -via V2 -resistance 1.00E-02
-set_layer_rc -via V3 -resistance 1.00E-02
-set_layer_rc -via V4 -resistance 1.00E-02
-set_layer_rc -via V5 -resistance 1.00E-02
-set_layer_rc -via V6 -resistance 1.00E-02
-set_layer_rc -via V7 -resistance 1.00E-02
-set_layer_rc -via V8 -resistance 1.00E-02
-set_layer_rc -via V9 -resistance 1.00E-02
+set_layer_rc -via V1 -resistance 1.72E-02
```
proppy, line 26: is that expected that the change is not consistant across layers?

proppy (approved):
just some nits and qs; can you remove the draft label if it's ready for review?

QuantamHD:
This looks good to me

dependency_support/org_theopenroadproject_asap7/rc_script.tcl
lpawelcz: Layers M1-9 are routing layers. Here are definitions of [scaled](https://github.com/The-OpenROAD-Project/asap7/blob/master/asap7sc7p5t_28/techlef_misc/asap7_tech_4x_201209.lef#L117-L132) and [non-scaled](https://github.com/The-OpenROAD-Project/asap7/blob/master/asap7sc7p5t_28/techlef_misc/asap7_tech_1x_201209.lef#L86-L101) M1 layers.

I assume that the layer thickness is the same in scaled and non-scaled variants and the routing wire placed on that layer is 4 times wider and longer in scaled variant than in non-scaled. In [LEF files](http://coriolis.lip6.fr/doc/lefdef/lefdefref/LEFSyntax.html#RoutingResistance), resistance is expressed in ohms per square of wire so this parameter does not depend on the changes (scaling) made to width and length of the wires on given layer.
My understanding is that if these resistances were to be specified in the LEF file, then those would have been the same for scaled and non-scaled variants.

However, in argument `-resistance`  for the command `set_layer_rc` the unit of resistance is ohm per length of default width wire. I'm not sure what is the interpretation of `default width` (perhaps the value specified in LEF tech file?) but in the docs for [set-wire-rc](https://openroad.readthedocs.io/en/latest/main/src/rsz/README.html#set-wire-rc) which handles `-resistance` similarly it is also stated that `The resistance and capacitance values are per length of wire, not per square or per square micron`.

I'm not sure if I understand that correctly but it looks reasonable to me that `ohm per length of default width wire` is not a universal unit and it should be scaled accordingly together with the width of the wires on given layer.

dependency_support/org_theopenroadproject_asap7/pdn_config.pdn
```diff
@@ -28,17 +28,24 @@ set_voltage_domain -name {CORE} -power {VDD} -ground {VSS}
...
+add_pdn_stripe -grid {top} -layer {M2} -width {0.018} -pitch {0.54} -offset {0} -followpins
+add_pdn_stripe -grid {top} -layer {M5} -width {0.12} -spacing {0.072} -pitch {11.88} -offset {0.300}
+add_pdn_stripe -grid {top} -layer {M6} -width {0.288} -spacing {0.096} -pitch {12} -offset {0.513}
 add_pdn_connect -grid {top} -layers {M1 M2}
 add_pdn_connect -grid {top} -layers {M2 M5}
 add_pdn_connect -grid {top} -layers {M5 M6}
-
-pdn::allow_repair_channels true
+####################################
+# macro grids
```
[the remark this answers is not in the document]
lpawelcz: This change is not required. In my test design it looks like it does not affect the result GDS file when it's removed.

dependency_support/org_theopenroadproject_asap7/pdn_config.pdn
lpawelcz: `-defer_connection` was removed in ORFS because this option is deprecated in newer `openroad`. However, the version used in `bazel_rules_hdl` still has it so it would be reasonable to keep that.

dependency_support/org_theopenroadproject_asap7/rc_script.tcl
lpawelcz: This would make sense if there are differences in thickness and widths of the via layers. I will check that

lpawelcz: So there is https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/commit/877e20089d684555ceb734b9354544e79fb467c8. I've also looked into layer definitions in PDK techfiles and the layers have different widths so it is expected that resistances will be different for each layer.

dependency_support/org_theopenroadproject_asap7/rc_script.tcl
lpawelcz: The most recent and relevant change to the original file was done in https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/pull/144. It mentions `scaling correlation` which probably refers to scaled and non-scaled variants of the PDK. The PR however changed only the capacitance while the resistance remained the same.
It is not clear to me if this is correct.
Later in https://github.com/The-OpenROAD-Project/asap7/issues/35 it was mentioned that `OpenROAD-flow-scripts` is the source of correct resistances and capacitances so maybe it is OK to use those.

@proppy please let me know what do you think about that.

lpawelcz:
Rebased after https://github.com/hdl/bazel_rules_hdl/pull/178 was merged

PR #187, opened 2023-09-18
Adding structured stat output.

mithro:
Use yosys to write a stats into a JSON file.


tcal-x (approved):
LGTM

mithro:
Should the bazel rules reference `STATS_JSON` in some way?

PR #188, opened 2023-09-18
Give don't use cells to Yosys.

mithro:


tcal-x:
Was this created using the copy tool or manually?  Do we care about proper attribution?

QuantamHD:
I wrote this, and I don't mind if it's not properly attributed.

mithro:
> Was this created using the copy tool or manually? Do we care about proper attribution?

This was created manually.

mithro:
Looks like we need to upgrade the version of `abc`?
```
ERROR: Command syntax error: Unknown option or option in arguments.
> abc -liberty bazel-out/k8-fastbuild/bin/external/com_google_skywater_pdk_sky130_fd_sc_hd/timing/sky130_fd_sc_hd__ff_100C_1v95.lib -dff -g aig -dont_use sky130_fd_sc_hd__probe_p_* -dont_use sky130_fd_sc_hd__probec_p_*
>                                                                                                                                               ^
ERROR: TCL interpreter returned an error: Yosys command produced an error
```

mithro:
I've upgraded both Yosys and ABC and the build runs fine locally now, will see what cloud build thinks.

QuantamHD (approved):

PR #191, opened 2023-09-18
Adding support for `synth.tcl` to read SystemVerilog files using Surelog.

mithro:


QuantamHD (approved):

tcal-x (approved):
LGTM

mithro:
The bazel rules should probably allow setting `USE_SURELOG_FRONTEND` somehow....

QuantamHD:
We don't have surelog in the upstream repo so we would probably need to add that as a prerequisite. I filed https://github.com/hdl/bazel_rules_hdl/issues/194 to keep track of it.

PR #196, opened 2023-09-20
Upgrade OpenROAD

mithro:
This pull request includes;
 * #197 
 * #186 
 * #201

Adds `ortools`.

PR #197, opened 2023-09-27
Update `rules_python`, and vendor requirements.bzl

ericastor:
The `rules_python` dependency has gotten out of date. We don't need any of the upgrades here, but since the way pip dependencies are managed has changed, this is blocking anyone using us from upgrading their own `rules_python` dependency.

To fix this, upgrade to the latest `rules_python` release, and update how pip dependencies are handled. Follows the recommended best-practices for vendoring the generated `requirements.bzl`.


mithro (commented):
FYI - This change is needed before I can land the OpenROAD update CL @ https://github.com/hdl/bazel_rules_hdl/pull/196

ericastor:
So - the tests run and pass locally for me. I'm not sure why the CI is failing, and can't seem to get the actual test log back out... If anyone knows how we can get the test log back out from the BuildBuddy run, that'd be EXTREMELY helpful.

mithro (approved):
It works!

PR #200, opened 2023-09-28
Export pdk/build_defs.bzl so that StandardCellInfo can be used (and tested) by XLS

tcal-x:
Not sure if I need to export `open_road_configuration` as well, or just the entry point?   The goal is to access  `do_not_use_cell_list`.   This is the only export I need internally.

### DIFF
--- a/pdk/BUILD
+++ b/pdk/BUILD
@@ -11,3 +11,5 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+
+exports_files(["build_defs.bzl"])


QuantamHD:
Bzl files don't require an export to use them in other bzl files

tcal-x:
> Bzl files don't require an export to use them in other bzl files

There is a meta-Bazel test `//xls/build_rules:xls_rules_bzl_parse_file_test` that does need this export.

This is the failure from XLS bazel testing (line breaks added for readability):

```
ERROR: /home/tcal/xls-export/xls/build_rules/BUILD:126:12: 
no such target '@rules_hdl//pdk:build_defs.bzl': 
target 'build_defs.bzl' not declared in package 'pdk' defined by
 /usr/local/google/home/tcal/.cache/bazel/_bazel_tcal/69bdc7c11d7de775bee1f8b708726bda/external/rules_hdl/pdk/BUILD;
however, a source file of this name exists.  
(Perhaps add 'exports_files(["build_defs.bzl"])' to pdk/BUILD?) 
and referenced by '//xls/build_rules:xls_codegen_rules_bzl'
```

QuantamHD:
Interesting. Didn't know that.

QuantamHD (approved):

PR #203, opened 2023-10-04
Use the floating bazel version on CI.

mithro:


QuantamHD (approved):

mithro:
**_Ideally_** I would like to run both the latest version of bazel and the oldest "supported" version of bazel.

PR #208, opened 2023-10-13
Add a line to synth.tcl to remove Yosys $print cells.

tcal-x:
Add a command to Yosys synthesis scripts to delete $print cells.
These cells would result in `always` blocks and $display tasks in the synthesized netlist;
some backend tools can't handle these.

### DIFF
--- a/synthesis/synth.tcl
+++ b/synthesis/synth.tcl
@@ -52,6 +52,9 @@ yosys proc -nomux
 yosys proc_mux
 yosys flatten
 
+# Remove $print cells.
+yosys delete {*/t:$print}
+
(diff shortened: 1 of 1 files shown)


synthesis/synth.tcl
```diff
@@ -52,6 +52,9 @@ yosys proc -nomux
 yosys proc_mux
 yosys flatten
 
+# Remove $print cells.
```
mithro, line 55: A description of **_why_** `$print` cells are being removed would be good.

tcal-x: Updated!

QuantamHD (approved):

PR #211, opened 2023-11-07
Fix ASAP7 for reworked upstream repo structure.

mithro:


dependency_support/org_theopenroadproject_asap7_pdk_r1p7/asap7.bzl
```diff
@@ -78,5 +82,10 @@ asap7_cell_library = rule(
             executable = True,
             cfg = "exec",
         ),
+        "_uncompress": attr.label(
+            default = Label("@7zip//:7za"),
```
QuantamHD, line 86: Do we have this tool imported inside Google?

QuantamHD (approved):

dependency_support/org_theopenroadproject_asap7_pdk_r1p7/asap7.bzl
mithro: Looks like it.

WORKSPACE
```diff
@@ -90,6 +90,17 @@ llvm_toolchain(
...
 
+maybe(
+    http_archive,
+    name = "rules_7zip",
+    strip_prefix = "rules_7zip-e00b15d3cb76b78ddc1c15e7426eb1d1b7ddaa3e",
+    urls = ["https://github.com/zaucy/rules_7zip/archive/e00b15d3cb76b78ddc1c15e7426eb1d1b7ddaa3e.zip"],
+    sha256 = "fd9e99f6ccb9e946755f9bc444abefbdd1eedb32c372c56dcacc7eb486aed178",
+)
+
+load("@rules_7zip//:setup.bzl", "setup_7zip")
```
proppy, line 101: maybe this should go into dependency support instead? so that downstream of this package don't have to do the same thing in their workspace?

mithro: I don't understand bazel enough to know when something should go into dependency_support verse WORKSPACE. @QuantamHD - Do you have any guidance?

PR #223, opened 2023-11-12
Load the `py_XXX` targets.

mithro:
Use the targets from `@rules_python//python:defs.bzl` rather than the native targets (which are being deprecated).

### DIFF
--- a/cocotb/BUILD
+++ b/cocotb/BUILD
@@ -12,6 +12,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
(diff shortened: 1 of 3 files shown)


mithro:
`grep -R 'py_' $(find -name BUILD\*) | sort | uniq` shows that all `BUILD` files that use `py_XXX` targets are loading the `@rules_python//python:defs.bzl` rules.

QuantamHD (approved):

QuantamHD:
LGTM

PR #227, opened 2023-11-20
Add `package(` rules to all `BUILD` files

mithro:
Make sure that all `BUILD` files have `package(` rules.


vivado/BUILD
```diff
@@ -1,3 +1,22 @@
+# Copyright 2023 Google LLC
```
QuantamHD, line 1: None of the vivado stuff was written by us. So maybe change to the bazel_rules_hdl authors or the like.

QuantamHD (approved):

vivado/BUILD
mithro: Done!

PR #238, opened 2023-12-05
Add another mirror location for bz2 tarball.

cdleary:


proppy:
Looks like there is a bunch of official mirrors there as well: https://sourceware.org/mirrors.html

QuantamHD (approved):

PR #241, opened 2023-12-10
Multi directory IP block fix

stridge-cruxml:
The ip repo path was getting overwritten when ip blocks were created in different directories.
Instead add all paths together in a list.


mithro (approved):
LGTM.

PR #242, opened 2023-12-11
Add `copts` to `verilator_cc_library`

hovind:
I don't know if this is the correct way to go about this, but I want to compile the `verilator_cc_library` with `-Os`, and this seemed like a simple way to achieve that goal.


mithro (approved):
LGTM. Will merge if CI goes green.

hovind:
> LGTM. Will merge if CI goes green.

Thanks for taking a look, and for all the time you and others have put into these rules. They're great!

PR #244, opened 2023-12-14
Add option to enable/disable `improve_placement` for P&R targets

mtdudek:
This adds new `enable_improve_placement` attr to `place_and_route` rule.
`improve_placement` pass breaks some of the designs,
so having option to  disable is beneficial for some users.
Default behavior remains unchanged.

First part of the #243 PR.


mithro (approved):
LGTM!

PR #245, opened 2023-12-14
P&R: Change database extension

mtdudek:
This PR changes database extension from.db to .odb.
OpenROAD gui file explorer only allows for .odb files to be loaded.
It doesn't affect GDS rules.

Second part of the #243


mithro (approved):
LGTM.

PR #247, opened 2023-12-15
place_and_route: detailed_routing: Add reports

lpawelcz:
It would be nice to have additional reports after detailed routing is finished.
This PR adds those in `detailed_routing` step. Report commands are similar to the ones in [global_routing](https://github.com/hdl/bazel_rules_hdl/blob/main/place_and_route/private/global_routing.bzl#L67-L72) step.
This is part of https://github.com/hdl/bazel_rules_hdl/pull/243 and https://github.com/google/xls/issues/1226

@mithro please take a look


mithro (approved):
LGTM

mithro:
Will merge on Monday, remind me if you don't see that happening.

PR #249, opened 2023-12-18
P&R: Run estimate_parasitics before clock repair

mtdudek:
This commit adds parasitic estimation after `clock_tree_synthesis` and
before `repair_clock_nets`. This pass is also run in the ORFS.

Third part of the #243

### DIFF
--- a/place_and_route/private/clock_tree_synthesis.bzl
+++ b/place_and_route/private/clock_tree_synthesis.bzl
@@ -47,6 +47,7 @@ def clock_tree_synthesis(ctx, open_road_info):
             sink_clustering_max_diameter = "-sink_clustering_max_diameter {}".format(ctx.attr.sink_clustering_max_diameter) if ctx.attr.sink_clustering_max_diameter else "",
         ),
         "set_propagated_clock [all_clocks]",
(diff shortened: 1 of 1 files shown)


place_and_route/private/clock_tree_synthesis.bzl
```diff
@@ -47,6 +47,7 @@ def clock_tree_synthesis(ctx, open_road_info):
             sink_clustering_max_diameter = "-sink_clustering_max_diameter {}".format(ctx.attr.sink_clustering_max_diameter) if ctx.attr.sink_clustering_max_diameter else "",
         ),
         "set_propagated_clock [all_clocks]",
+        "estimate_parasitics -placement",
         "repair_clock_nets",
         "estimate_parasitics -placement",
```
mithro, line 52: Is `estimate_parasitics -placement` still needed after `repair_clock_nets` now?

mtdudek: I think yes, as `repair_clock_nets` may add extra buffer cells.

mithro (approved):
LGTM.

PR #250, opened 2023-12-18
P&R: Add set_propagated_clock before detail_route

mtdudek:
This commit adds `set_propagated_clock` pass to the detail route script.
This modification can affect final design, so extra tests may be required.
It brings bazel_rules_hdl closer to the ORFS detailed_routing implementation.

Fourth part of the #243


mithro (approved):
LGTM

PR #251, opened 2023-12-18
P&R: Move `remove_buffers` pass to floorplan stage

mtdudek:
Move `remove_buffers` pass to run just after floorplan initialization
It `remove_buffers` should only remove clock/data buffers
added by synthesis process. Changing order may affect designs.

Fifth part of the #243


mithro (approved):
LGTM!

PR #252, opened 2023-12-18
Synthesis: Insert tie cells during synthesis

mtdudek:
This commit adds following changes to the synthesis script:
* defined all `x` values as `0`s
* split multi-bit nets to single bit ones
* run extra optimization after `x` values are resolved.
* map constant 0/1s to tap cells during synthesis. These steps are similar to the ORFS synthesis.

Sixth part of the #243


mithro (approved):
LGTM.

Issue #239, opened 2023-12-08
Comparing with OpenROAD-flow-scripts


mtdudek:
Recently I've been comparing Verilog to GDS flow between bazel_rules_hdl and OpenROAD-flow-scripts.

I've been testing this flows using common verilog design and
ASAP7 7.5T rev 28 RVT cells library.
The tested design is generated from DSLX language using XLS.

I've tried using bazel_rules_hdl to get GDS for that design, with clock period of 2.5ns.
Design is failing on the global routing stage with error message
`[ERROR GRT-0118] Routing congestion too high. Check the congestion heatmap in the GUI.`
P&R parameters:
```
clock_period = "2500.0",
core_padding_microns = 0,
die_width_microns = 200,
die_height_microns = 200,
placement_density = "0.16",
```

Then I tried running the same source verilog using ORFS.
With the same clock constraints ORFS easily achieved 2.5 ns,
it was able to reach timings < 1ns using RVT cells.
ORFS parameters:
```
set clk_name  core_clock
set clk_port_name clk
set clk_period 2500
set clk_io_pct 0.01

set clk_port [get_ports $clk_port_name]

create_clock -name $clk_name -period $clk_period $clk_port

set non_clock_inputs [lsearch -inline -all -not -exact [all_inputs] $clk_port]

set_input_delay  [expr $clk_period * $clk_io_pct] -clock $clk_name $non_clock_inputs
set_output_delay [expr $clk_period * $clk_io_pct] -clock $clk_name [all_outputs]
```

I later tried running bazel_rules_hdl synthesized Verilog through 
ORFS P&R stages. It failed, as ORFS expects tie cells to be
already placed in the synthesized Verilog.
I've modified synthesis script, to use tie cells from openroad_pdk provider
and that was enough to get synthesized Verilog through ORFS,
it also achieved clock period < 1ns using RVT cells.

I've also tested starting ORFS from floorplan stage using flororplan generated 
by bazel_rules_hdl. This run gave following error message:
`[ERROR GRT-0119] Routing congestion too high. Check the congestion heatmap in the GUI`
It is exactly the same error message. 119 means that extra information was being logged to
a file, while 118 means that there is no extra file.

I checked which version of ASAP7  ORFS use. It's 7.5t rev 28 with some extra cells defined.
Extra cells can be removed from the picture. Bazel_rules_hdl synthesized Verilog
only uses standard ASAP7 RVT cells and running ORFS P&R on it gave good timing results.

[...]

PR #253, opened 2023-12-18
Synthesis: Add fa/ha extract and techmap pass

mtdudek:
This commit adds extra `extract_fa` Yosys pass combined with fa/ha techmapping.
It runs only when user specifies $fa/$ha to DPK mapping.
This commit also adds new `adder_mapping` attr to the `synthesize_rtl` rule.
This file provides Yosys with $fa/$ha blackboxes.
This change may affect existing designs, as extra `opt` pass is being run.

Seventh part of the #243


mithro (approved):
This change looks good but nothing seems to use it?

mithro:
As described in https://github.com/hdl/bazel_rules_hdl/issues/239 I'm going to wait a for the other changes to bake for a bit before landing these.

mtdudek:
I'm not sure if I should attach verilog files that are PDK and yosys specific,
so I left it up to the user to provide such mapping.

mithro (commented):
Can we provide ASAP7 & SKY130 configurations?

Maybe at https://github.com/hdl/bazel_rules_hdl/tree/main/dependency_support/org_theopenroadproject_asap7_pdk_r1p7 and https://github.com/hdl/bazel_rules_hdl/tree/main/dependency_support/com_google_skywater_pdk/sky130_fd_sc_hd ?

mtdudek:
I can add them to this folders.
Should I modify PDK provider to also include ha/fa mapping files?

mithro:
> Should I modify PDK provider to also include ha/fa mapping files?

I assume that is what is needed to get us into the same set up as `OpenROAD-flow-scripts`?

mtdudek:
I've added fa/ha mapping files from ORFS and I've modified `open_road_configuration`
to provide fa/ha mapping file. Doing so I've also rebased branch onto the newest main.

mithro:
LGTM!

PR #254, opened 2023-12-18
Synthesis: Use ABC script from ORFS

mtdudek:
This commit adds abc script from ORFS as a default abc script.
User can provide custom script using `abc_script` attr in the `synthesize_rtl` rule.
This is not a default yosys abc script, so it may cause some issues with existing flows.

Eighth part of the #243

### DIFF
--- a/synthesis/BUILD
+++ b/synthesis/BUILD
@@ -22,6 +22,7 @@ package(
 )
 
 exports_files(["synth.tcl"])
+exports_files(["abc.script"])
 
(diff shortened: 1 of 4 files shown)


mithro (approved):
LGTM

mithro:
As described in https://github.com/hdl/bazel_rules_hdl/issues/239 I'm going to wait a for the other changes to bake for a bit before landing these.

PR #255, opened 2023-12-19
Don't rely on escapes for spaces in logging tcl command

allight:
A yosys log command used an escape to print a trailing space. This could cause issues if the trailing space was removed by code-cleanup tools. Instead quote the logged string.


allight:
@proppy @QuantamHD PTAL

QuantamHD (approved):

PR #266, opened 2024-01-10
Adding `-ffp-contract=off` flag.

mithro:
This is needed for floating point stability.

The global placement depends on the output of the FFT library. Without `-ffp-contract=off` the current FFT library produces different results dependent on if the compiler is targeting an x86 architecture which has the "FMA instruction".

The FMA instruction omits the rounding done in the multiplication instruction, making the calculations produce different results for some input values

See https://kristerw.github.io/2021/11/09/fp-contract/

See also OpenROAD change @
https://github.com/The-OpenROAD-Project/OpenROAD/pull/4518

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -13,7 +13,9 @@
 # limitations under the License.
 
 build --cxxopt "-std=c++17"
+build --cxxopt "-ffp-contract=off"
 build --host_cxxopt "-std=c++17"
+build --host_cxxopt "-ffp-contract=off"
 build --crosstool_top=@llvm_toolchain//:toolchain
 
 build:ciremotebuild --crosstool_top=@llvm_toolchain//:toolchain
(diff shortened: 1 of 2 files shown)


QuantamHD (approved):

QuantamHD:
Should this go in the bazelrc as a global option as well?

mithro:
> Should this go in the bazelrc as a global option as well?

Potentially, I don't know how that works?

QuantamHD:
You can add flags to the bazel command line by default so `--copt="-ffp-contract=off"`. 

https://github.com/hdl/bazel_rules_hdl/blob/main/.bazelrc#L17

So something like `build --copt="-ffp-contract=off"`

mithro:
> You can add flags to the bazel command line by default so `--copt="-ffp-contract=off"`.
> 
> https://github.com/hdl/bazel_rules_hdl/blob/main/.bazelrc#L17
> 
> So something like `build --copt="-ffp-contract=off"`

Like this -> https://github.com/hdl/bazel_rules_hdl/pull/266/files#diff-544556920c45b42cbfe40159b082ce8af6bd929e492d076769226265f215832fR16 ?

QuantamHD:
Looks good!

PR #274, opened 2024-01-15
verilator: Output generated headers

hovind:
The generated headers are tracked through the `CcInfo` provider. This pull request adds the generated header files to the `files` field of the `DefaultInfo` provider.

In our case, we want to use the generated headers and verilated library outside of bazel. In other words, we want a the generated headers to be exposed as output files for a target.

I'm not sure if `verilated.h` and its dependencies should be packed in here as well, so I left it out for this PR.

Please don't hesitate to reach out!

### DIFF
--- a/verilator/defs.bzl
+++ b/verilator/defs.bzl
@@ -157,7 +157,7 @@ def _verilator_cc_library(ctx):
     defines = ["VM_TRACE"] if ctx.attr.trace else []
     deps = [ctx.attr._verilator_lib, ctx.attr._zlib, ctx.attr._verilator_svdpi]
 
-    return cc_compile_and_link_static_library(
+    [default, cc] = cc_compile_and_link_static_library(
         ctx,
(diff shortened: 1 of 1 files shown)


abrisco (changes requested):
Why not use `OutputGroupInfo`? I would not expect the generated headers to be the thing I find in `DefaultInfo.files` They feel more like an implementation detail.

hovind:
> Why not use `OutputGroupInfo`? I would not expect the generated headers to be the thing I find in `DefaultInfo.files` They feel more like an implementation detail.

Using an `OutputGroupInfo` provider sounds much better, thanks!

Also thanks for taking a look, and for all the improvements to these rules.

hovind:
> Why not use `OutputGroupInfo`? I would not expect the generated headers to be the thing I find in `DefaultInfo.files` They feel more like an implementation detail.

I pushed a new proposal, all advice is much appreciated.

PR #277, opened 2024-01-30
Adds C++20 support for LEMON

QuantamHD:


grebe:
This seems to build correctly against my c++20 build, thanks!

PR #280, opened 2024-02-15
Fix flop count in synth.tcl

ericastor:
Fix one glob used to match flops in the default synth.tcl flow; previously excluded flops in some PDKs.


QuantamHD (approved):

QuantamHD:
Waiting on CI otherwise LGTM! Thanks @ericastor

PR #288, opened 2024-03-02
Updated Verilator to not reply on system python

abrisco:
In my remote executor image I explicitly don't want `python`/`python3` to appear on PATH and to instead rely on toolchains defined in Bazel to correctly match an interpreter. I noticed there were some genrules when building Verilator that were relying on this behavior. I've replaced the genrules used by Verilator with proper rules and `py_binary` targets to make builds more hermetic.


QuantamHD:
Running CI. Thanks for the contribution

QuantamHD:
Can you resolve the merge conflict. Otherwise looks good

abrisco:
@QuantamHD rebased!

abrisco:
@QuantamHD Looks like CI is passing 😃

QuantamHD (approved):

QuantamHD (approved):

PR #292, opened 2024-03-03
Upgrade Verilator to 5.022

abrisco:
Upgraded Verilator [from 4.224 to 5.022](https://github.com/verilator/verilator/blob/v5.022/Changes#L11-L695).


abrisco:
@QuantamHD Looks like CI is passing 😃

QuantamHD (approved):

PR #299, opened 2024-03-16
Consolidated verilator copy-tree actions

abrisco:
This also avoids reliance on the behavior of `mkdir` and `cp` which may differ on different systems.


QuantamHD (approved):

QuantamHD:
Nice! Thanks for the contribution

PR #306, opened 2024-03-28
Remove uses of ctx.resolve_tools(...).

tjgq:
This API will be removed from a future Bazel version.

Instead, the `executable` or `tools` argument to `ctx.actions.run` and `ctx.actions.run_shell` should be used.


tjgq:
@QuantamHD Can you retrigger the pipeline? There was a typo.

tjgq:
@QuantamHD another try?

I'm having trouble getting this to build locally, but based on the error message, I believe I wasn't passing all input files to the swig action, and I've fixed that now.

QuantamHD:
Added you as a triager. That should allow the workflow to run automatically

tjgq:
@QuantamHD Looks like it's good to go now.

QuantamHD:
Thanks very much @tjgq!

QuantamHD (approved):

PR #313, opened 2024-04-03
Fix namespace of PdkInfoProto to better reflect the intent.

hzeller:
Namespaces for protocol buffers should not be based on path-elements but primarily on what they represent.
This proto has nothing to do with bazel nor rules, just hdl.


hzeller:
Let's fix the name while it is not too late and more people depend on it.

mikesinouye (approved):

PR #319, opened 2024-04-15
Refine proto buffer describing PPA results.

hzeller:
This standardizes units for power (watts), time (picoseconds), and length units (um or um2) and breaks out more information that might be available from the EDA tools.

Compared to the original proto, there are so many changes, that this is a new protocol buffer, based on the old one. So to not clash with the original, this uses a new name PowerPerformanceAreaProto; the old proto needed a small change in the name of a type (because we really want to use the name 'Power' in the new one with a different definition), but it is fully binary compatible to the one before.

TL;DR:
  * old: hdl.ppa.PerformancePowerAreaProto
  * new: hdl.ppa.PowerPerformanceAreaProto

Next step will be to adapt the bazel rules currently generating the old proto to create the new one.

Once this new one is established, the old one can be deprecated and removed quickly after last use.

### DIFF
--- a/synthesis/power_performance_area.proto
+++ b/synthesis/power_performance_area.proto
@@ -1,0 +1,166 @@
+syntax = "proto2";
+
+package hdl.ppa;
+
+// A word about units: The units used by the EDA-tools internally might vary
+// depending on context or as provided by the PDK.
+//
+// This proto is meant to always use standardized units so that it is
+// easy to compare; the program that is closest to the source of the data
+// reading logfiles and reports can convert these metrics once;
+// downstream we don't have to worry about it anymore.
+//
+// We're using
+//  - power units:  always watts (that reads easiest in scientific notation).
+//  - time units:   always picoseconds. This allows sufficient resolution to
+//                  store them as integer values.
+//  - length units: always micrometers (or micrometers squared for area).
+
+// A protocol buffer describing performance metrics for a particular design.
+message PowerPerformanceAreaProto {
+  optional Power power = 1;
+  optional Performance performance = 2;
+  optional Area area = 3;
+
+  // Information about the operating corner.
+  optional string cell_library = 4;
+  optional float operating_voltage = 5;
+
+  // Arbitrary string that might be used to describe the
+  // job that created this data, e.g. command line options.
+  optional string build_info = 10;
+}
+
+message Area {
+  // Length units here are abbreviated in ASCII-compatible
+  // way in the identifier suffixes.
+  //  `um`  - Micrometer (length)
+  //  `um2` - square Micrometer (area)
+
+  // The whole die area.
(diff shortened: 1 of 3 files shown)


synthesis/power_performance_area.proto
```diff
@@ -0,0 +1,165 @@
...
+  optional int32 num_combinational = 11;
+
+  // The number of non-timing buffers.
+  optional int32 num_buffers = 12;
+
+  // The number of timing buffers.
+  optional int32 num_timing_buffers = 13;
+
+  // The number of flops.
+  optional int32 num_flops = 14;
```
mikesinouye, line 76: I am reversing what I said earlier, perhaps "sequential" would be better instead of "flops". There are some sequential cells like synchronizers that are not flops, which may cause the total to add to total_cells. Sequential seems more generic, so we could be less precise so things still add up to the total.

synthesis/power_performance_area.proto
```diff
@@ -0,0 +1,165 @@
...
+  map<string, float> celltype_fractions = 20;
+}
+
+message Performance {
+  // Desired clock frequency this PPA metric was generated for.
+  // It influences the choice of area-impacting optimization, so this
+  // helps to understand area size in that context.
+  optional int32 clock_period_ps = 1;
+
+  // Longest path in picoseconds between any two flop-boundaries.
```
mikesinouye, line 91: I would omit the "two flop boundaries" part, as I think we would still want to report an in-to-FF and a FF-to-out delay as the critical path, if it is.

synthesis/power_performance_area.proto
```diff
@@ -0,0 +1,165 @@
...
+  optional int32 num_hold_violations = 11;
+  optional int32 num_slew_violations = 12;
+  optional int32 num_fanout_violations = 13;
+  optional int32 num_capacitance_violations = 14;
+
+  // The number of cells in the longest topological path.
+  optional int32 longest_topological_path = 15;
+
+  // Number of cells in the critical path
+  optional int32 critical_path_cells = 16;
```
mikesinouye, line 136: How is this different than above? Does longest path mean in-to-out? If so, I would specify that int he comment, as a I was assuming "longest_topological_path" was a constrained path.

mikesinouye (changes requested):

synthesis/power_performance_area.proto
hzeller: done.

synthesis/power_performance_area.proto
hzeller: Done.

synthesis/power_performance_area.proto
hzeller: Clarified in comment that the criticial path one is based on timing.

mikesinouye (approved):

PR #322, opened 2024-05-05
Add `data` attribute to `verilog_library`

abrisco:
This change includes the following:
- `verilog_library` limit `hdrs` to `.vh` and `.svh` files.
- `verilog_library` limit `srcs` to `.v` and `.sv` files.
- `verilog_library` gain a `data` attribute for all other inputs.


QuantamHD:
Unfortunately, I won't be able to merge this. The VerilogInfo provider mirrors an implementation inside Google that I don't control, but fundamentally need to maintain compatibility with. The full implementation has a bunch of functionality related to tagging verilog targets [SIM, ELAB, SYNTH, etc.] so that you can dynamically build a DAG for different targets based on tags for different purposes. 

I personally wouldn't have done it that way, but unfortunately what's done is done. Here's a snippet of the additional things kept around in the upstream version.

```
    return struct(
        srcs = tuple(defs + srcs),
        hdrs = tuple(hdrs),
        defs = tuple(defs),
        data = tuple(data),
        experimental = struct(exclusion_list_files = tuple(experimental_exclusion_list_files)),
        flists = tuple(flists),
        label = label,
        legacy = struct(cc = tuple(legacy_cc)),
        plis = depset(plis + dpis),
        tags = tuple(tags),
        deps = tuple(deps),
    )
```

abrisco:
> Unfortunately, I won't be able to merge this. The VerilogInfo provider mirrors an implementation inside Google that I don't control, but fundamentally need to maintain compatibility with. The full implementation has a bunch of functionality related to tagging verilog targets [SIM, ELAB, SYNTH, etc.] so that you can dynamically build a DAG for different targets based on tags for different purposes.
> 
> I personally wouldn't have done it that way, but unfortunately what's done is done. Here's a snippet of the additional things kept around in the upstream version.
> 
> ```
>     return struct(
>         srcs = tuple(defs + srcs),
>         hdrs = tuple(hdrs),
>         defs = tuple(defs),
>         data = tuple(data),
>         experimental = struct(exclusion_list_files = tuple(experimental_exclusion_list_files)),
>         flists = tuple(flists),
>         label = label,
>         legacy = struct(cc = tuple(legacy_cc)),
>         plis = depset(plis + dpis),
>         tags = tuple(tags),
>         deps = tuple(deps),
>     )
> ```

Is the only issue the change from `tuple` to `depset`? I can add a global setting for that or leave it as is for now and at least benefit from a much smaller list of things to iterate over (splitting `srcs` and `data`).

QuantamHD:
That seems to be the only difference. The real limitation is that downstream consumers of VerilogInfo can't have a breaking API change. So theoretically, no changes to synthesis/build_defs.bzl

abrisco:
> That seems to be the only difference. The real limitation is that downstream consumers of VerilogInfo can't have a breaking API change. So theoretically, no changes to synthesis/build_defs.bzl

I can make those changes. But that being said, would you be opposed to the introduction of a new provider? or the introduction of a build setting that would support toggling potentially breaking changes?

QuantamHD:
LGTM

QuantamHD (approved):

PR #324, opened 2024-05-05
Add `defs.bzl` for `//synthesis` package

abrisco:


QuantamHD:
What's the overall goal with defs.bzl change? Looks harmless overall, but just trying to get better context.

At Google, I don't frequently see the defs.bzl redirection.

abrisco:
> What's the overall goal with defs.bzl change? Looks harmless overall, but just trying to get better context.
> 
> At Google, I don't frequently see the defs.bzl redirection.

This is to try and reduce the mental overhead of remembering where to load things from or needing to constantly look this up. The two patterns I've found to be most common are `<package>/<package>.bzl` or `<package>/defs.bzl` for generally where to load content from. Jumping around from `defs.bzl`, `providers.bzl`, `build_defs.bzl`, etc in the repo currently feels like an unnecessary mental tax and my hope here is to make that better without introducing a breaking change by moving files.

I think it'd be fine to reject this if you don't prefer `defs.bzl` but it would be nice if module names were predictable across `rules_hdl`.

QuantamHD:
That's totally fine with me. Just wanted to know the motivation.

QuantamHD (approved):

PR #334, opened 2024-06-26
Fixes to allow downstream project to set a different Python toolchain.

mikex-oss:
Currently, if a downstream project attempts to use a custom Python toolchain, passes it to bazel_rules_hdl `init` as instructed, and the toolchain doesn't happen to be named exactly as configured for bazel_rules_hdl itself, the initialization will fail with:

> Error in path: Unable to load package for @python39_x86_64-unknown-linux-gnu//:bin/python3: The repository '@python39_x86_64-unknown-linux-gnu' could not be resolved: Repository '@python39_x86_64-unknown-linux-gnu' is not defined

These changes clean up Python toolchain and vendoring of requirements.bzl.

It also pins a Bazel version for Bazelisk (official Bazel version management), updates rules_python, and removes the dependency on pybind11.


WORKSPACE
```diff
@@ -47,12 +47,22 @@ bazel_skylib_workspace()
 maybe(
     http_archive,
     name = "rules_python",
-    sha256 = "5868e73107a8e85d8f323806e60cad7283f34b32163ea6ff1020cf27abef6036",
-    strip_prefix = "rules_python-0.25.0",
-    url = "https://github.com/bazelbuild/rules_python/releases/download/0.25.0/rules_python-0.25.0.tar.gz",
+    sha256 = "e3f1cc7a04d9b09635afb3130731ed82b5f58eadc8233d4efb59944d92ffc06f",
+    strip_prefix = "rules_python-0.33.2",
+    url = "https://github.com/bazelbuild/rules_python/releases/download/0.33.2/rules_python-0.33.2.tar.gz",
+    #sha256 = "5868e73107a8e85d8f323806e60cad7283f34b32163ea6ff1020cf27abef6036",
```
ericastor, line 53: We should probably remove these commented lines...

mikex-oss: Oops, thanks.

ericastor (approved):

PR #338, opened 2024-07-23
Add *Machine.cc files that are included by Machine.cc.

grebe:
OpenRoad includes different `Machine*.cc` files on different platforms, and only Linux was previously included. This change makes all files available to be included.

### DIFF
--- a/dependency_support/org_theopenroadproject/bundled.BUILD.bazel
+++ b/dependency_support/org_theopenroadproject/bundled.BUILD.bazel
@@ -1428,7 +1428,10 @@ cc_library(
     hdrs = glob(
         include = ["src/sta/include/sta/*.hh"],
     ) + [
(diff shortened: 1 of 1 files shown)


QuantamHD:
This is going to lead to ODR violations. You'll need to bazel select them based on the @platforms// target.

grebe:
I don't think it should because it's stuffed into `hdrs`. Confusingly, it looks like they're `#include`ing `.cc` files.

I don't disagree that selecting them would be better, though.

QuantamHD:
I think technically these should be included in the textual_hdrs, and not the regular hdrs since they aren't meant to be compiled.

QuantamHD:
I think this was fixed by https://github.com/hdl/bazel_rules_hdl/pull/373

PR #339, opened 2024-07-25
Fixes out of memory error with larger RTL blocks

QuantamHD:
The autoname pass seems to cause issues with larger rtl blocks. Since the autoname pass really only assits with net names, and to be honest not that much. We can just remove it for now.


QuantamHD:
@ericastor FYI for internal bug

mikesinouye (approved):

PR #342, opened 2024-08-05
Fix a bug in the pin placement script feature of place_and_route.

mikesinouye:
@mikex-oss discovered a bug in which a custom pin placement script is only added to the bazel input flist if there is also a tapcell TCL provided by the open_road_configuration tech library setup.


QuantamHD (approved):

QuantamHD:
Nice catch

mikesinouye:
Thanks @mikex-oss for the find and fix!

PR #345, opened 2024-08-14
Bump rules_proto pin

abrisco:
In [rules_proto v6.0.0](https://github.com/bazelbuild/rules_proto/releases/tag/6.0.0) a breaking change was introduced that moved a public facing macro. This change updates the version of this repository and accounts for the breaking change.

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -151,14 +151,6 @@ maybe(
     url = "https://github.com/bazelbuild/rules_proto/releases/download/6.0.2/rules_proto-6.0.2.tar.gz",
 )
 
-load("@rules_proto//proto:repositories.bzl", "rules_proto_dependencies")
-
-rules_proto_dependencies()
-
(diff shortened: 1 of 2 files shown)


WORKSPACE
```diff
@@ -131,16 +131,34 @@ load("@rules_7zip//:setup.bzl", "setup_7zip")
...
 )
 
+load("@rules_proto//proto:repositories.bzl", "rules_proto_dependencies")
+
+rules_proto_dependencies()
+
+load("@rules_proto//proto:toolchains.bzl", "rules_proto_toolchains")
+
+rules_proto_toolchains()
+
```
QuantamHD, line 161: Can we move these back to the init? Otherwise we'll need to instruct users they'll need a viable proto toolchain instatnitanted in their workspace

abrisco: Done!

QuantamHD (approved):

PR #346, opened 2024-08-20
bump net_zlib to v1.3.1

PeterAaser:
v 1.2.11 causes warnings which are fixed here:
https://github.com/madler/zlib/commit/e9d5486e6635141f589e110fd789648aa08e9544

Without this we get errors along the lines of
```
external/net_zlib/trees.c:379:20: warning: a function definition without a prototype is deprecated in all versions of C and is not supported in C23 [-Wdeprecated-non-prototype]
```

### DIFF
--- a/dependency_support/net_zlib/net_zlib.bzl
+++ b/dependency_support/net_zlib/net_zlib.bzl
@@ -21,10 +21,10 @@ def net_zlib():
     maybe(
         http_archive,
         name = "net_zlib",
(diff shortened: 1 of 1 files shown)


PeterAaser:
I just realized I ran tests in the wrong repo, so it's likely that this PR is broken. I'm not at a computer until tomorrow so I cannot verify. Terribly sorry, this is not a great first impression...

PeterAaser:
Things came up. I'm not able to actually run these tests so I'm closing this until I'm back at the computer.

PR #347, opened 2024-08-23
Adds ability to use standard cells in user provided RTL

QuantamHD:
This change also includes a new `verilog_defines` parameter to synthesize_rtl which allows users to set defines that will be used when parsing rtl

### DIFF
--- a/synthesis/build_defs.bzl
+++ b/synthesis/build_defs.bzl
@@ -140,6 +140,8 @@ def _synthesize_design_impl(ctx):
     if ctx.attr.verilog_defines:
         for define in ctx.attr.verilog_defines:
             args.add(define)
+
+    standard_cell_black_box = ctx.actions.declare_file("{}_stdcells_blackbox.v".format(ctx.attr.name))
     script_env_files = {
         "ABC_SCRIPT": abc_script,
         "ADDITIONAL_LIBERTIES": additional_liberty_files,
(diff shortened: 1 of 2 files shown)


synthesis/synth.tcl
```diff
@@ -12,6 +12,16 @@
...
+# create liberty blackbox so cells can be manually created.
+foreach lib $all_liberties {
+  read_liberty -lib -overwrite $lib
+}
+
+write_verilog -blackboxes $::env(STANDARD_CELL_BLACK_BOX)
+
 # read design
 set srcs_flist_path $::env(FLIST)
 set srcs_flist_file [open $srcs_flist_path "r"]
```
mikesinouye, line 27: Using this approach, do we not need to re-read the blockbox file?

synthesis/build_defs.bzl
```diff
@@ -137,6 +137,11 @@ def _synthesize_design_impl(ctx):
         for dont_use_pattern in or_config.do_not_use_cell_list:
             dont_use_args += " -dont_use {} ".format(dont_use_pattern)
 
+    if ctx.attr.verilog_defines:
+        for define in ctx.attr.verilog_defines:
+            args.add(define)
+
+    standard_cell_black_box = ctx.actions.declare_file("{}_stdcells_blackbox.v".format(ctx.attr.name))
```
mikesinouye, line 144: Also how does this work if the action both generates and consumes the file? Do we need to list it?

synthesis/synth.tcl
QuantamHD: Good catch. Messed up the copy and paste

synthesis/build_defs.bzl
QuantamHD: Also another miscopy. I added it to the output list of the action.

mikesinouye (approved):

PR #350, opened 2024-09-04
Add new timing endpoint breakdown fields to the ppa proto.

mikesinouye:


synthesis/power_performance_area.proto
```diff
@@ -118,16 +118,26 @@ message Performance {
...
-  // If negative, new data arrives too early for the DFF to ingest
-  // the previous one reliably.
+  // A negative value indicates that new data arrives too early for the
+  // endpoint to properly sample the previous value. A zero value indicates
+  // that timing is met - positive values are atypical but can be used to report
+  // margin. These values are the worst of in2reg, reg2reg, reg2out, and in2out.
   optional sint32 hold_wns_ps = 5;
   optional sint64 hold_tns_ps = 6;
 
+  // Additional timing breakdowns for in2reg, reg2reg, reg2out, and in2out.
```
QuantamHD, line 135: Maybe worth spelling out in and out mean block input to first register and last register to block output.

mikesinouye: Good point, added those.

QuantamHD (approved):

PR #356, opened 2024-10-01
dependency_support: bump protobuf, absl and ortools

proppy:


ericastor:
Looks like there's an issue - probably with the patch we're trying to apply from OR-Tools. Some sort of sequencing issue?

proppy:
> 
Looks like there's an issue - probably with the patch we're trying to apply from OR-Tools. Some sort of sequencing issue?

Needed to bump absl as well.

mikesinouye (approved):
Thank you!

PR #357, opened 2024-10-01
Set repair_tns to 100% when running repair_timing.

grebe:
Be explicit so we don't rely on any potentially surprising default value.


mikesinouye (approved):

mikesinouye:
Thanks!

PR #366, opened 2024-11-07
Use fewer threads when running openroad.

grebe:
Don't overwhelm remote executors (which may be running multiple actions) with too many threads. Just use one unless the job is required to be run locally, in which case we expect an action will often have the entire system to itself.

### DIFF
--- a/place_and_route/open_road.bzl
+++ b/place_and_route/open_road.bzl
@@ -245,6 +245,14 @@ def openroad_command(ctx, commands, input_db = None, step_name = None, inputs = 
         ),
     )
 
+    # Don't overwhelm remote executors (which may be running multiple actions)
+    # with too many threads. Just use one unless the job is required to be run
+    # locally, in which case we expect an action will often have the entire
+    # system to itself.
+    num_threads = "1"
+    if "no-remote-exec" in execution_requirements:
+        num_threads = "max"
+
     ctx.actions.run(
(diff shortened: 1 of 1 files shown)


place_and_route/open_road.bzl
```diff
@@ -245,14 +245,22 @@ def openroad_command(ctx, commands, input_db = None, step_name = None, inputs =
         ),
     )
 
+    # Don't overwhelm remote executors (which may be running multiple actions)
+    # with too many threads. Just use one unless the job is required to be run
+    # locally, in which case we expect an action will often have the entire
+    # system to itself.
+    num_threads = "1"
+    if "no-remote-exec" in execution_requirements:
```
QuantamHD, line 253: Will this trigger if you run with --strategy=OpenROAD=local?

grebe: It looks like setting the strategy still keeps num_threads=1 :/

Might be better to have num_threads be a parameter to override. WDYT?

QuantamHD: Could we check the opposite? i.e. is this running remotely?

hzeller:
What happened to this PR ?

grebe:
I think the intermediate- to long-running stuff that was hitting this shouldn't be run via remote executor, and there are better ways to make OR respect the remote executors constraints, so I'll close.

PR #372, opened 2024-12-12
Remove the unused implicit attribute `_cc_toolchains`.

mikesinouye:
CC toolchains are gathered from the line:
```
toolchains = use_cc_toolchains(),
```

### DIFF
--- a/cloudbuild.yaml
+++ b/cloudbuild.yaml
@@ -34,7 +34,7 @@ steps:
     path: /root/.ssh
 
 # Run the build
-- name: gcr.io/cloud-builders/bazel:6.4.0
+- name: gcr.io/cloud-builders/bazel:7.4.1
   entrypoint: 'bash'
   args:
   - '-c'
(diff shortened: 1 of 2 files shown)


katre:
Can your project upgrade to Bazel 7.4.1? That should resolve the failure, and is much more up-to-date than the current vevrsion of 6.x.

mikesinouye:
While that is the best fix, briefly experimenting with bazel 7 seems like it requires a cascade of dependency upgrades/fixes. At this point, the time is likely better spent upgrading to bzlmod instead. @QuantamHD should we stick with bazel6 for now, resolve our internal issue with a patch, and plan to spend time for a proper bazel 8 + bzlmod migration sometime next year?

QuantamHD:
Might want to just go with a patch internally, and move to bzlmod

mikesinouye:
Sounds good, let's go with that.

PR #374, opened 2025-01-15
Bump the libffi version to newest, and update rules to build on MacOS

JoshVarga:


QuantamHD (approved):

QuantamHD:
If CI passes will merge.

PR #376, opened 2025-01-27
Don't load cc_proto_library from rules_cc: not there anymore.

hzeller:
Since 6.x, the cc_proto_library() is part of default bazel rules https://bazel.build/versions/6.5.0/reference/be/c-cpp

And removed as of 0.1.0 from rules_cc:
https://github.com/bazelbuild/rules_cc/releases/tag/0.1.0

making the load call fail if used in a context with newer rules_cc.

### DIFF
--- a/pdk/proto/BUILD
+++ b/pdk/proto/BUILD
@@ -12,7 +12,6 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-load("@rules_cc//cc:defs.bzl", "cc_proto_library")
 load("@rules_proto//proto:defs.bzl", "proto_library")
 load("@rules_python//python:proto.bzl", "py_proto_library")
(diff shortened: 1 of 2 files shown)


hzeller:
buildifier is drunk (or too old)

What it means is to load it from `load("@com_google_protobuf//bazel:cc_proto_library.bzl", "cc_proto_library")`, but that is not available in the requested protobuf version. So this change is the best that is compatible with all worlds currently.

QuantamHD:
Is there a way to disable buildifier on those lines with a comment? Otherwise this step will always fail

hzeller:
Looks like there is!

hzeller:
So, ready to merge ?

mikesinouye (approved):

PR #379, opened 2025-02-16
Avoid calling `bc`, to minimize binaries needed in bazel environment.

hzeller:
In minimal installations, bc might not be available. We already process the data with awk, so we have everything needed to do the calculations.


mikesinouye (approved):

mikesinouye:
Thanks @hzeller

PR #383, opened 2025-02-25
Delete non-deterministic outputs from VerilatorCompile actions

UebelAndre:
This change introduces a means to avoid non-deterministic outputs from `VerilatorCompile` actions by adding a process wrapper that deletes unwanted files after verilator runs.

Currently the only file deleted is [`{prefix}__verFiles.dat`](https://verilator.org/guide/latest/files.html) which contains timings for each file generated. This information is not used any downstream actions for the rule so is assumed to be generally unused.

This change provides backward compatibility through the `--@rules_hdl//verilator/settings:avoid_nondeterministic_outputs` flag which users can set to `false` to disable this new behavior.


UebelAndre:
> So this is caused by the use of declare_directory. Is it possible to explicitly declare all the outputs from a verilator compile action?

It is but I don’t have confidence in the names of the outputs. I had looked a while back and understood the file names to be something you couldn’t predict with Bazel and I feel this is also supported by the existence of the copy action for secs and headers. I would love to declare files and delete the extra action if possible but I don’t think it is.

QuantamHD (approved):

QuantamHD:
Sounds good.

PR #385, opened 2025-02-28
Update rules flex/bison to latest.

hzeller:

PR #386, opened 2025-02-28
Update Verilator to 5.0.34

hzeller:


mikesinouye (approved):

hzeller:
@abrisco looks good ?

Also, afterwards, can you have a look if there is anything here that resolves the problem in #385  ? Looks like you have mostly maintained the `dependency_support/verilator/` path in recent times.

hzeller:
Looks like with the latest changes in bazel, this also works.

PR #392, opened 2025-03-21
Upgrades abc

QuantamHD:


mikesinouye:
@QuantamHD This change may be breaking xls-oss. They are seeing the error: `external/edu_berkeley_abc/src/proof/ssw/sswPart.c:36:10: fatal error: 'atomic' file not found`

It's possibly related to the `copts = ["-x c++",],` change. To unblock them, I am going to roll back - let's discuss when you're back.

PR #393, opened 2025-03-21
Adds cudd

QuantamHD:


hzeller:
Neat, this looks like something the guy who works on https://github.com/The-OpenROAD-Project-staging/OpenROAD/pull/4 might be interested in :smiley: 

Would you like to make it a bazel  central registry component https://registry.bazel.build/ ?
It is fairly simple to convert your changes here into a component there; I've done it for two projects already and it is pretty straightforward. And future compatible.

maliberty:
@hzeller I think it would be great to make it into a central component.  This code is under no development so it shouldn't be any work to maintain.  You might want to use https://github.com/The-OpenROAD-Project/cudd as the repo to draw from.

PR #394, opened 2025-03-24
dependency_support: bump klayout to 0.29.12

ericastor:


ericastor:
This appears to pass on the test systems... but it doesn't actually pass on my local machine. The issue appears unrelated; I may reach out to ask for help here.

mikex-oss (approved):

mikesinouye:
Thank you for the fix.

PR #401, opened 2025-03-28
pseudo_configure: use `sed` instead of `perl`.

hzeller:
Perl is quite a big sledgehammer for the nut, and less likely to be available than `sed`.


hzeller:
Tested: Ran

```
bazel build @org_gnu_readline//:config_h
```

And the generated `bazel-bin/external/org_gnu_readline/readline/config.h` before and after were the same.

mikesinouye (approved):
Thanks!

PR #412, opened 2025-06-20
update to swig 4.1.0 (matching OpenROAD)

maliberty:


QuantamHD:
Clone and run `bazel build @org_swig//...`

PR #414, opened 2025-07-15
Refresh readline to 8.3

hzeller:
The old version, 8.0, creates warnings on modern compilers.


hzeller:
sigh, looks like this also would require updating the ancient magic. Will see if I have time on the weekend for that.

PR #418, opened 2025-08-14
Fix place_and_route benchmark metrics extraction

ericastor:
The metrics extraction logic seems to have had a few bugs; some may be longstanding, but others may have been rot.

The overall effect was that WNS and TNS were incorrectly reported as 0, and that (non-timing-repair) buffer area would be reported as equal to the buffer count. These are now fixed.


mikesinouye:
I think the issue here is that the OSS version of openroad used in this repo is very old, and the syntax used here is compatible with that version. We have not imported this repo directly in a while due to the skew.

We are very close to being able to bump the version of openroad used here - bazel build is very close to complete in openroad main.

PR #421, opened 2025-09-29
Exclude ncurses/report_hashing.c as it is a test with a main function

maliberty:


maliberty:
I don't understand what failed.

QuantamHD (approved):

PR #435, opened 2026-04-05
Add missing bzl_library fix buildifier warnings

edcote:
This PR adds missing bzl_library targets and establishes proper dependency chains for Starlark files in the pdk, place_and_route, and synthesis packages.


edcote:
context: b/378942548

edcote:
@mikesinouye do you need to aprove the /gcbrun?

PR #439, opened 2026-06-25
Add bzl_libraries for some bzl files.

allight:
This allows users to see these files for things like stardoc.

#vibe-coded


allight:
@QuantamHD @mikesinouye PTAL

allight:
This is needed in order to make https://github.com/google/xls/pull/4468 work because stardoc needs everything to be in bzl_libraries, including across dependencies.

mikesinouye (approved):

PR #442, opened 2026-07-05
Fix invalid escape (backslash space) in a Python file.

hzeller:
This creates a lot of noisy warning messages when compiling XLS, so let's fix it at the root.
(unfortunately, the skywater github repo is archived, so need to apply as patch here).


mikesinouye:
Thanks Henner

mikesinouye (approved):

Issue #440, opened 2026-07-03
Discrepancy between Yosys synthesis logs and log parsing logic inside `benchmark_synth` target


magancarz:
I've noticed a few discrepancies between logs produced by Yosys synthesis and the log parsing logic inside `benchmark_synth` target:
- [cdd2fcc8](https://github.com/hdl/bazel_rules_hdl/commit/cdd2fcc8edf738c55ba923027ed814a9b9ab34f1) adds a [`-T` flag](https://github.com/YosysHQ/yosys/blob/8f07a0d8404f63349d8d3111217b73c9eafbd667/kernel/driver.cc#L262) to `synthesize_rtl` rule in `synthesis/build_defs.bzl` which causes timing statistics to be omitted in logs
- [c3fb8ff5](https://github.com/hdl/bazel_rules_hdl/commit/c3fb8ff546155f806aec54cae3cf7764ec444559) changes grep pattern to `'End of script.*time:'` which is incorrect for the currently used version of Yosys. Correct pattern would be `'End of script.*CPU:.user'` as visible in timing statistics logs [here](https://github.com/YosysHQ/yosys/blob/8f07a0d8404f63349d8d3111217b73c9eafbd667/kernel/driver.cc#L701)

This causes `benchmark_synth` targets to return exit code 1 and fail. It can either be solved by removing the `-T` flag from synthesis and fixing the grep pattern or by not extracting timing statistics.

mikesinouye:
Either solution is fine by me, which does XLS / Antmicro prefer? Are those stats useful?

magancarz:
Opened a pull request with changes fixing this discrepancy in #443.

magancarz:
Closing as [#443](https://github.com/hdl/bazel_rules_hdl/pull/443) was merged.

PR #443, opened 2026-07-07
Fix discrepancy between Yosys synthesis logs and log parsing logic inside `benchmark_synth` target

magancarz:
This pull request fixes discrepancy between Yosys synthesis logs and log parsing logic inside `benchmark_synth` target described in #440.


magancarz:
I've rebased on latest main so this pull request contains [95f0ce1c](https://github.com/hdl/bazel_rules_hdl/commit/95f0ce1c2a821759b05dd3e385f1a1e7891b875e) which adds mirror for gperf dependency. I've also experienced problems with mirror for glpk used by or-tools and local override allowed me to run all tests.

mikesinouye (approved):

PR #445, opened 2026-07-29
cocotb: fix build_args parameters passing

mczyz-antmicro:
This PR fixes 2 issues with cocotb rules:
* If you passed build arguments through the build_args, e.g. `-Wno-Y`, `--Wno-X`, the script would interpret it as an argument to the python wrapper and fail with an error. Now you can pass the dash-arguments to the simulator correctly.
* I added a second test to check if "-Wall" is correctly passed to verilator, which uncovered that when you try to build multiple tests in the same directory, e.g.
`bazel test //cocotb/tests:*`
you get an error 
`ERROR: file 'cocotb/tests/cocotb_runner.sh' is generated by these conflicting actions:`
so I patched the name of the cocotb_runner.sh script to be unique by appending target name.

### DIFF
--- a/cocotb/cocotb.bzl
+++ b/cocotb/cocotb.bzl
@@ -27,6 +27,9 @@ def _list_to_argstring(data, argname, attr = None, operation = None):
         result += " {}".format(elem)
     return result
 
+def _repeated_argstring(data, argname):
+    return "".join([" --{}={}".format(argname, value) for value in data])
+
 def _dict_to_argstring(data, argname):
     result = " --{}".format(argname) if data else ""
     for key, value in data.items():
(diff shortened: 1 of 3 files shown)


mczyz-antmicro:
Currently fails on missing verilator executable:
```
/builder/home/.cache/bazel/_bazel_root/eab0d61a99b6696edb3d2aff87b585e8/sandbox/linux-sandbox/13130/execroot/_main/bazel-out/k8-fastbuild/bin/cocotb/tests/cocotb_counter_verilator_cocotb_runner.sh.runfiles/_main/cocotb/cocotb_wrapper.py:26: UserWarning: Python runners and associated APIs are an experimental feature and subject to change.
  from cocotb.runner import get_runner, check_results_file
ERROR: verilator executable not found!
```

Given how the rules are structured, may I ask the maintainers to install verilator in the image to unblock this PR? Long-term it would be better to create a dependency between the cocotb rules and the verilator toolchain, but I think it would be reasonable to do it in a follow-up. Alternatively we could drop the counter_verilator test from this PR and readd it together with refactoring the toolchains, WDYT?

mikesinouye:
> Alternatively we could drop the counter_verilator test from this PR and readd it together with refactoring the toolchains, WDYT?

I think generally for bazel build this is the best approach, we'd like to minimize hidden dependencies as we can. Would you mind doing that as a follow up?

mczyz-antmicro:
Updated PR:
* dropped the failing verilator test
* added build_args to the existing icarus test

@mikesinouye not at all, will try to follow up with the toolchain update

mikesinouye (approved):

PR #449, opened 2026-09-19
Fix missing C++ standard library includes for newer LLVM toolchains.

allight:
- Update `dependency_support/at_clifford_yosys/yosys.patch` to explicitly include `<stdlib.h>` in `libs/subcircuit/subcircuit.cc` (for `free()`) and `libs/ezsat/ezsat.h` (for `abort()`).
- Add `bazel_dep(name = "rules_cc_autoconf", version = "0.24.0")` in `MODULE.bazel` to pick up the upstream `<iterator>` include fix for `std::back_inserter`.
- Patch `rules_bison` via `single_version_override` to include `<cstdlib>` in `src/bazel_runfiles.cc` for `getenv()`.

This is motivated by the LLVM toolchain upgrade in XLS where stricter libc++ headers no longer transitively include `<cstdlib>` or `<iterator>`.

Change generated using Google Antigravity


allight:
@mikesinouye PTAL

allight:
Oh Sorry @mikesinouye I think I killed the /gcbrun by fixing a lint error

🤦

mikex-oss (approved):

mikex-oss:
Just checking, does the lockfile not need to be updated?

allight:
Done. Whoops was mostly building this from outside of the project so forgot that wouldn't update the lockfile.

mikex-oss (approved):

